Issue #14, opened 2018-10-18
TypeError: Cannot read property 'contentType' of undefined


mouafa:
I think the issue is because the for loop `for (const contentType of contentTypes)` is setting both the types and the nodes in the same loop. 

<img width="1000" alt="error" src="https://user-images.githubusercontent.com/1658799/47165126-a911b600-d2f9-11e8-88d2-d4a6c6974182.png">

PR #15, opened 2018-10-18
fix(souce-contentful): define all types before creating nodes

mouafa:
this should fix issue #14


hjvedvik:
Thanks! Works great :)

PR #21, opened 2018-10-28
Updated for image support

AmGarera:
This adds the better_featured_image endpoint to GraphQL and the only requirement is that the Wordpress install has the Better REST API Featured Images Plugin https://wordpress.org/plugins/better-rest-api-featured-images/ To add native support for images without this plugin "?_embed" would need to appended to the end of each URL.

### DIFF
--- a/packages/source-wordpress/index.js
+++ b/packages/source-wordpress/index.js
@@ -114,7 +114,8 @@ class WordPressSource {
           slug: post.slug,
           fields: {
             content: post.content ? post.content.rendered : '',
-            excerpt: post.excerpt ? post.excerpt.rendered : ''
+            excerpt: post.excerpt ? post.excerpt.rendered : '',
+            betterFeaturedImage: post.better_featured_image ? post.better_featured_image.source_url : ''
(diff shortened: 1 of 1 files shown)


tomtev:
Hi @AmGarera, could you check if this is available in the fields node without this fix? Everything that's getting added to the WP rest api should already be available in the GraphQL layer

hjvedvik:
Thanks for the PR! I think it is better to solve this without requiring plugins in WordPress. WordPress also sends tons of data on each post if `?_embed` is add to the url. So I made the plugin fetch attachments instead which is referenced from the posts through `node.refs.featuredMedia`.

Also the Gridsome plugin API has changed, so you are getting some conflicts. Please merge inn master and check if you want to add more fields that are relevant for the front end instead. But you can also check if the `featuredMedia` schema field [here](https://github.com/gridsome/gridsome/blob/b938f72ce2c0f9404ccc2b131f68f9845df0a67a/packages/source-wordpress/index.js#L78) will work as you expect.

PR #25, opened 2018-11-02
Graphql image

EldoranDev:
This PR should add support for images in Frontmatter.


gridsome/lib/graphql/schema/types/image.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  return true
+}
+
+exports.imageType = {
+  type: GraphQLJSON,
+  args: {},
+  resolve: async (fields, { }, context, { fieldName }) => {
+    return await context.queue.add(fields[fieldName]);;
```
hjvedvik: The `g-image` component only need `src`, `srcset`, `sizes`, `size` and `dataUri` from the queue results. So you can probably return those properties.

gridsome/lib/graphql/schema/types/image.js
```diff
@@ -0,0 +1,31 @@
...
+  if (path.basename(value) === value) {
+    return false
+  }
+
+  return true
+}
+
+exports.imageType = {
+  type: GraphQLJSON,
+  args: {},
```
hjvedvik: The queue can take `width` and `quality` options. Would be nice to have them as arguments? They can be simple int types for now.

hjvedvik:
I'll merge this in as it is now. We can add arguments etc. later. Thanks for the work you've put into this :)

PR #33, opened 2018-11-09
@gridsome/source-datocms plugin

stefanoverna:
Alright @hjvedvik, this works with gridsome version on master! The only thing that's not working is refs with multiple nodes, but I think that's related to this https://github.com/gridsome/gridsome/blob/master/gridsome/lib/graphql/schema/nodes/createType.js#L127-L130

Let me know if everything's fine with this PR!

### DIFF
--- a/packages/source-datocms/index.js
+++ b/packages/source-datocms/index.js
@@ -68,7 +68,7 @@ class DatoCmsSource {
 
       const contentType = {
         typeName: makeTypeName(itemType.name),
-        route: routes[itemType.name] || `${slugify(itemType.name)}/:slug`,
+        route: routes[itemType.name] || `/${slugify(itemType.name)}/:slug`,
       }
 
       const collection = addContentType(contentType)
--- a/packages/source-datocms/index.js
+++ b/packages/source-datocms/index.js
@@ -1,0 +1,137 @@
+const { SiteClient, Loader } = require('datocms-client')
+const { camelize } = require('humps')
+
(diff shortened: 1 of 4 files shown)


packages/source-datocms/index.js
```diff
@@ -0,0 +1,137 @@
...
+    for (const itemType of itemsRepo.itemTypes) {
+      const { titleField, fields } = itemType
+
+      const slugField = fields.find(({ fieldType }) => fieldType === 'slug')
+
+      cache[itemType.id] = { titleField, slugField }
+
+      const contentType = {
+        typeName: makeTypeName(itemType.name),
+        route: routes[itemType.name] || `${slugify(itemType.name)}/:slug`,
```
hjvedvik: Can you prefix the fallback route with a `/`?

stefanoverna:
Sure!

Il giorno ven 9 nov 2018, 13:10 Hans-Jørgen Vedvik <notifications@github.com>
ha scritto:

hjvedvik:
You are right, references with multiple node types isn't ready yet, but I want to look into it as soon as possible. I've only added one minor comment to the code that should be fixed, but everything else looks good :)

Devportobello:
Hi @stefanoverna 

Can you show me example of using with "page-query" for "single record" and "multiple records" ?

For single record, i'm facing an "A DatoCmsHome with id undefined was not found" error while no error when using the graphql.datocms.com :/

PR #34, opened 2018-11-09
extend the configuration for faker plugin

valentinvieriu:
allow to modify the route for faker plugin


hjvedvik:
Thanks! That is a useful option :) Just one minor thing, can you move the fallback route to default options instead?

valentinvieriu:
@hjvedvik updated the file as requested

Issue #40, opened 2018-11-15
Support for Drupal


matt-e-king:
Hello gridsome maintainers and community,

I work for the the University of Arizona, College of Agriculture and we primarily develop within VueJS and Drupal. We recently created a PWA with a "headless" Drupal 8 instance as our backend. Drupal 8 exposes a lot more RESTful endpoints and with the help of a jsonapi Drupal module, we were able to standardize this even further. 

Do you know if there is a roadmap for building out @gridsome/source-drupal (Drupal 8 support)? Or do you know any contributors in the community who might be starting this effort? If not, this is something we will be pursuing internally and would love to share our progress.

Thanks!

hjvedvik:
Drupal is on our todo-list, but we haven't had the opportunity to look into it yet. No one is working on it as far as I know. And personally, I have only experience with Drupal 7, so it would be great if you want to create a source plugin for Drupal 8 :)

You can follow the [How to contribute](https://github.com/gridsome/gridsome#how-to-contribute) guide and create a PR. Then we can follow the progress and answer any questions you might have. Some APIs are still under development, so we can improve core with any necessary APIs or bugfixes if you run into problems during development. You can read about the [Data Store API](https://gridsome.org/docs/data-store-api) or look at the source code for the other source plugins to get started.

matt-e-king:
Thanks, @hjvedvik, I will start that process now.

matt-e-king:
https://github.com/gridsome/gridsome/pull/41

PR #41, opened 2018-11-15
@gridsome/source-drupal

matt-e-king:
https://github.com/gridsome/gridsome/issues/40


matt-e-king:
@tomtev @hjvedvik This will only support Drupal 8, any thoughts on if we should name space the directory or file structure to reflect that? Or should it just be mentioned in the README?

matt-e-king:
@hjvedvik I've committed what I think is a good start for Drupal Integration. The approach is not specific to one type of Drupal REST methodology, it really is just a generic wrapper where the end user can provide an array of endpoints and associate the response from each end point with a GraphQL Type (I explain this in more detail in the README). 

The more I think about it, this can be used as more of a catch-all plugin? It doesn't even need to be specific to Drupal. Once you get a chance to look over this I'd be curious what you think.

Thanks!

hjvedvik:
Good job! :) I'll take a closer look at it as soon as possible. And yes, this could almost just be a source plugin for any Rest API endpoint which respons with a list of resources. But I think we can add more Drupal specific functionalities sometime later in this source plugin maybe? So we can look at a `@gridsome/souce-rest-api` plugin or something later.

matt-e-king:
I agree, I want it to feel more Drupal specific as well. I chatted with one of my colleagues who is our in-house Drupal expert about how I could make this plugin feel more "out of the box" with Drupal and he basically reiterated similar findings. 

The closest thing to "out of the box" would be **Views**, but Drupal 7 doesn't even come equipped with the right modules (Serializer, Web RESTful Services) to convert them to REST endpoints. Regardless, there isn't a "standard" set of Views, it still requires developer intervention to set them up. But the developer has full control over the payload, so any Drupal "standard" can be washed away, and all you end up with is just an endpoint, with a potentially non-standard response.

Drupal 8 now comes equipped with RESTful Services, but they are for only dealing with one node at a time - there are no GET requests for all nodes or all nodes by a filter. To make this plugin more specific to Drupal we could tack on support for each of the following popular modules:

- [JSON:API](https://www.drupal.org/project/jsonapi)
- [Services](https://www.drupal.org/project/services) (which I believe is basically what Drupal 8 offers, but more extensible?)
- [Drupal GraphQL](https://www.drupal.org/project/graphql)

And if we supported each of these popular methods, we could provide a configuration option in gridsome.config.js for which module the user is using in Drupal. Just a thought...

patrickcate:
The JSON API module is expected to be included in Drupal 8 core next year. 

https://wimleers.com/blog/state-of-jsonapi-2018-07

hjvedvik:
I've checked out the JSON:API module and I think it could work very well for Gridsome. I created a test site and installed the module without any further configuration. The module comes with several useful endpoints out of the box. I was able to get a list of node types and also list all nodes of that type. Custom node types are also automatically getting endpoints for listing nodes.

The response also comes with useful relationship info which Gridsome could use to reference other types in the GraphQL schema. Especially the relationships which starts with `field_*`.

Maybe we should just create this plugin for that module?

The test API is here: http://dev-gridsome.pantheonsite.io/jsonapi

matt-e-king:
@hjvedvik Sounds good! Yes, I just recently created a PWA with a headless Drupal using JSON:API and was able to create a generic API handler for any content type pretty easily. So I'm definitely on board for this! @patrickcate Glad to hear it is going to be a part of Drupal 8 core - I didn't know how "accepted" it was by the community.

I'm thinking we could still have`views` (currently called `types`)  as an option in the configuration so the user can still configured additional endpoints that JSON:API doesn't satisfy.

I'll take an attempt at this over the next week...

tanc:
This is good work so far! I'm about to embark on a feasibility study for our little Drupal agency to see whether Gridsome (or other front ends) would work well for us and in what situations.

Reading the code of this source plugin actually makes it easier to understand how the loading happens in Gridsome before I've even installed it. The README is handy too.

I'll report back findings as I get deeper into things.

tanc:
Also to note that the [Contenta](https://github.com/contentacms/contenta_jsonapi) distribution is leading the charge at the moment as an API first Drupal content platform. I believe it uses JSON API to provide endpoints.

matt-e-king:
@tanc that's great to hear! We are actually in the process of creating a new version of this Drupal plugin to work primarily with JSON:API. This current version is more of a "catch all" and not specific to Drupal.

I'm closing in on uploading a first pass which will hit the root JSON:API endpoint `/jsonapi/` which returns a schema generated by the Drupal JSON:API module, and then use that schema to fetch the rest of the resources.

As a proof of concept, I currently have all `node` based entities working and all `taxonomy_term` based entities working. I will hopefully have this committed soon!

tanc:
Excellent! Looking forward to it

matt-e-king:
Type up notes and quick explanation now...

matt-e-king:
Just pushed an update for this plugin to work directly with JSON:API. I will begin working on a README today (hopefully). Here some quick notes:

This plugin now requires the following Drupal modules:
- [JSON:API](https://www.drupal.org/project/jsonapi)
- Serializer

To make things easier, you should also install:
- [RESTUI](https://www.drupal.org/project/restui)
- RESTful Services

### Quick run-through:
This is how it currently works: 
- The plugin options that are currently supported are:
-- `baseUrl` self explanatory
-- `entities` this is an array of the different entities you want this plugin to create content out of

```
module.exports = {
  plugins: [
    {
      use: '@gridsome/source-drupal',
      options: {
        baseUrl: '<drupal admin url>',
        entities: ['node', 'taxonomy_term', 'file', 'user']
      }
    }
  ]
}
```

**entities:**
The only current supported entities are `['node', 'taxonomy_term', 'file', 'user']`
- node - drupal "content types"
- taxonomy_term - drupal terms and vocabulary (like tags and categories)
- file - all files uploaded through drupal admin
- user - self explanatory
You will see how this naming convention matches up with the response of `/jsonapi`

**WARNING: I believe `taxonomy_term`,  `file`, and `user` jsonapi endpoints will return errors until those permissions are explicitly set in Drupal under `/admin/people/permissions`. I can provide more details in the README.**

cont'd
- On initialization, it fetches `/jsonapi` from the `baseUrl`.
-- This endpoint returns the schema/manifest of all the endpoints made available by JSON:API
-- It then normalizes that response data into something more manageable like:
```
[
       {
         entityType: 'node',
         entityName: 'article',
         type: 'node--article',  // json api type
         url: '<jsonapi endpoint>'
       },
       {
         entityType: 'taxonomy_term',
         entityName: 'tags',
         type: 'taxonomy_term--tags', // json api type
         url: '<jsonapi endpoint>'
       },
       ...
]
```

Notice how `entityType` is the prefix of `type`. We use this `entityType` value to verify against the `options.entities` array passed in by the user in the `gridsome.config.js`. The `/jsonapi` response keys all the urls by the jsonapi type for Drupal content types, there could be multiple `node--` urls.

**i.e.**
- node--article
- node--page
- node--\<custom content type\>

cont'd
- It then loops through that and checks if:
-- the `entityType` was passed in the `options.entities` array from user's `gridsome.config.js`
-- checks an internal instance property of `supportedEntities` (which is just an object mapping which references imported js files from `/src/entities/*.js`)
- If those checks pass, it then initializes a new class instance for each supported `entityType`
-- i.e. `node--article` which is a built in content type will initialize `/src/entities/Nodes.js`
-- i.e. `node--page` which is a built in content type will initialize `/src/entities/Nodes.js`
-- i.e. `file--file` will initialize `/src/entities/Files.js`

Each of these `/src/entities/*.js` classes pretty much do the same thing:
- Fetches the data using the url provides in the schema from above
-- if the response returns a `next` url, it paginates through requesting all the content
- Creates a GraphQL Type (`addContentType`) named after the JSON:API type
-- `node--article` becomes `NodeArticles`
-- `taxonomy-term--tags` becomes `TaxonomyTermTags`
-- `file--file` becomes `FileFiles`
-- Some of those GraphQL type names can be a little cumbersome but this way it helps namespace. Because imagine if a user created a Drupal content type called "files" then there would be `node--files` which would create a GraphQL type called `NodeFiles` - this can be distinguished against `FileFiles`. (I'm sure there is possible clean up here in the future)
- Anyways...
- Then loops through all the data in the response and runs `addNode` against the GraphQL Type
- Determines "customFields"
- Then loops through the `relationship` object provided in the response and keeps it for later :) 

### Relationships:
Once this has all been done, the main `DrupalSource` instance then calls `buildRelationships` on each instance that was created. This has to happen AFTER all GraphQL Type are created, otherwise it would throw errors. 

`buildRelationships()` verifies that the content type exists (`getContentType`). This whole area still feels a little messy because there is a lot of transforming and normalizing. The field name of the relationships are determines by what they are "keyed" as in the jsonapi response - open up the GraphQL explorer to see. (i.e. user relation is called `RevisionUID`, I'm hoping to clean that up later)

### GraphQL queries examples:
```
query Articles {
  allNodeArticles {
    edges {
      node {
        id,
        title,
        path,
        revisionUid {
          name
        },
        fieldTags {
          id,
          title
        },
        fieldImage {
          id,
          filename
        }
      }
    }
  }
}
```
```
query Files {
  allFileFiles (perPage: 1000) {
    edges {
      node {
        id,
        title,
        path
      }
    },
    totalCount
  }
}
```
```
query Tags {
  allTaxonomyTermTags {
    edges {
      node {
        id,
        title,
        path
      }
    }
  }
}
```
```
query Users {
  allUserUsers {
    edges {
      node {
        id,
        title,
        path,
        userPicture {
          filename
        }
      }
    }
  }
}
```

That's it! FYI - I haven't done strenuous testing against all different Drupal field types and scenarios yet. 

### Roadmap:
Going forward, here is what I think can be done:
- [ ] Create/publish Gridsome project example
- [ ] relationships come with important meta data that isn't being merge yet
- [ ] test all field types in Drupal in order to appropriately transform data for GraphQL
- [ ] open up more options for the user (custom path for each entity type, additional headers for the jsonapi request, anything else?)
- [ ] better inline documetation
- [ ] unit tests

Should this package be renamed to `source-drupal-jsonapi`?

packages/source-drupal/src/DrupalSource.js
```diff
@@ -0,0 +1,201 @@
...
+
+
+class DrupalSource {
+  // defaultOptions merged with this.options in App.vue
+  static defaultOptions() {
+    return {
+      baseUrl: '',
+      views: [], // deprecated
+      entities: [],
+      format: 'json'
```
tanc:
```suggestion
      format: 'json',
      apiUrl: 'jsonapi',
```

packages/source-drupal/src/DrupalSource.js
```diff
@@ -0,0 +1,201 @@
...
+
+      await this.processEntities()
+    } catch (error) {
+      let { message } = error
+      throw new Error(message)
+    }
+  }
+
+  async fetchJsonApiSchema() {
+    const { baseUrl } = this.options
```
tanc:
```suggestion
    const { baseUrl, apiUrl } = this.options
```

packages/source-drupal/src/DrupalSource.js
```diff
@@ -0,0 +1,201 @@
...
+
+    if (!baseUrl) throw new Error('baseUrl is required in gridsome.config.js')
+
+    try {
+      let response = await axios.get(`${baseUrl}jsonapi/`)
```
tanc:
```suggestion
      let response = await axios.get(`${baseUrl}${apiUrl}/`)
```

packages/source-drupal/src/DrupalSource.js
```diff
@@ -0,0 +1,201 @@
...
+   * 
+   * Once each entityType is processed, loop through again and buildRelationships()
+   */
+  async processEntities() {
+    await Promise.all(
+      this.apiSchema.map((api) => {
+        let { entities } = this.options
+        let { entityType, entityName, type, url } = api
+
+        if (this.supportedEntities[entityType] && entities.includes(entityType)) {
```
tanc:
```suggestion
        if (this.supportedEntities[entities[entityType]] && entityType in entities) {
```

packages/source-drupal/src/DrupalSource.js
```diff
@@ -0,0 +1,201 @@
...
+          // creating an instance of the entity class, see ./entities/*
+          this.entities[type] = new this.supportedEntities[entityType](this, { entityType, entityName, type, url })
```
tanc:
```suggestion
          this.entities[type] = new this.supportedEntities[entities[entityType]](this, { entityType, entityName, type, url })
```

packages/source-drupal/src/DrupalSource.js
hjvedvik: This would be a nice option. Maybe `apiBase` is a better option name since it's not a full URL?

packages/source-drupal/src/entities/TaxonomyTerms.js
```diff
@@ -0,0 +1,71 @@
...
+    const config = this.createGraphQlType()
+    this.graphQlContentType = addContentType(config)
+
+    for (let item of this.response) {
+      let {
+        id,
+        attributes: {
+          name,
+          created,
+          description: {
```
tanc: Can throw errors if `description` is null, as in this JSON:

```
{
  "type": "categories",
  "id": "1f808af2-bd6d-4576-a1df-5a74eb343545",
  "attributes": {
    "internalId": 31,
    "status": true,
    "name": "Accompaniment",
    "description": null,
    "weight": 0,
    ...
  }
}
```

packages/source-drupal/src/entities/Entity.js
```diff
@@ -0,0 +1,203 @@
...
+
+  get getType() {
+    return this.type
+  }
+
+  get getTypeName() {
+    return this.createTypeName(this.type)
+  }
+
+  createTypeName(text = '') {
```
hjvedvik: Other source plugins have a `typeName` option which is prefixed all types the plugin creates. This makes it easier to use Gridsome with other sources without naming collisions, and also have multiple Drupal sources. Should we do the same for this plugin? The types would then be like `DrupalNodeArticle`. And we can skip the name if it equals the entity type, so `user--user` will become `DrupalUser`. What do you think?

The store has a helper function for creating the type names. It doesn't add the `s` suffix, but we can easily add it for all source plugins later if that is desired (before 1.0 release).

```js
store.makeTypeName('node--article') >> DrupalNodeArticle
```

tanc:
I've just set up Contenta again, this time without the default content and the schema looks different, more like default JSONAPI schema:
```
{
  "data": [],
  "links": {
    "self": "http:\/\/agileback.lndo.site\/api",
    "block--block": "http:\/\/agileback.lndo.site\/api\/block\/block",
    "comment_type--comment_type": "http:\/\/agileback.lndo.site\/api\/comment_type\/comment_type",
    "consumer--consumer": "http:\/\/agileback.lndo.site\/api\/consumer\/consumer",
    "editor--editor": "http:\/\/agileback.lndo.site\/api\/editor\/editor",
    "entity_browser--entity_browser": "http:\/\/agileback.lndo.site\/api\/entity_browser\/entity_browser",
    "field_storage_config--field_storage_config": "http:\/\/agileback.lndo.site\/api\/field_storage_config\/field_storage_config",
    "field_config--field_config": "http:\/\/agileback.lndo.site\/api\/field_config\/field_config",
    "file--file": "http:\/\/agileback.lndo.site\/api\/file\/file",
    "filter_format--filter_format": "http:\/\/agileback.lndo.site\/api\/filter_format\/filter_format",
    "image_style--image_style": "http:\/\/agileback.lndo.site\/api\/image_style\/image_style",
    "jsonapi_resource_config--jsonapi_resource_config": "http:\/\/agileback.lndo.site\/api\/jsonapi_resource_config\/jsonapi_resource_config",
    "configurable_language--configurable_language": "http:\/\/agileback.lndo.site\/api\/configurable_language\/configurable_language",
    "language_content_settings--language_content_settings": "http:\/\/agileback.lndo.site\/api\/language_content_settings\/language_content_settings",
    "media--image": "http:\/\/agileback.lndo.site\/api\/media\/image",
    "media_type--media_type": "http:\/\/agileback.lndo.site\/api\/media_type\/media_type",
    "node_type--node_type": "http:\/\/agileback.lndo.site\/api\/node_type\/node_type",
    "node--person": "http:\/\/agileback.lndo.site\/api\/node\/person",
    "node--blog_post": "http:\/\/agileback.lndo.site\/api\/node\/blog_post",
    "paragraphs_type--paragraphs_type": "http:\/\/agileback.lndo.site\/api\/paragraphs_type\/paragraphs_type",
    "paragraph--image": "http:\/\/agileback.lndo.site\/api\/paragraph\/image",
    "paragraph--image_with_text": "http:\/\/agileback.lndo.site\/api\/paragraph\/image_with_text",
    "paragraph--text_block": "http:\/\/agileback.lndo.site\/api\/paragraph\/text_block",
    "rdf_mapping--rdf_mapping": "http:\/\/agileback.lndo.site\/api\/rdf_mapping\/rdf_mapping",
    "responsive_image_style--responsive_image_style": "http:\/\/agileback.lndo.site\/api\/responsive_image_style\/responsive_image_style",
    "rest_resource_config--rest_resource_config": "http:\/\/agileback.lndo.site\/api\/rest_resource_config\/rest_resource_config",
    "oauth2_token_type--oauth2_token_type": "http:\/\/agileback.lndo.site\/api\/oauth2_token_type\/oauth2_token_type",
    "menu--menu": "http:\/\/agileback.lndo.site\/api\/menu\/menu",
    "action--action": "http:\/\/agileback.lndo.site\/api\/action\/action",
    "taxonomy_vocabulary--taxonomy_vocabulary": "http:\/\/agileback.lndo.site\/api\/taxonomy_vocabulary\/taxonomy_vocabulary",
    "user--user": "http:\/\/agileback.lndo.site\/api\/user\/user",
    "user_role--user_role": "http:\/\/agileback.lndo.site\/api\/user_role\/user_role",
    "view--view": "http:\/\/agileback.lndo.site\/api\/view\/view",
    "menu_link_content--menu_link_content": "http:\/\/agileback.lndo.site\/api\/menu_link_content\/menu_link_content",
    "base_field_override--base_field_override": "http:\/\/agileback.lndo.site\/api\/base_field_override\/base_field_override",
    "date_format--date_format": "http:\/\/agileback.lndo.site\/api\/date_format\/date_format",
    "entity_form_mode--entity_form_mode": "http:\/\/agileback.lndo.site\/api\/entity_form_mode\/entity_form_mode",
    "entity_view_mode--entity_view_mode": "http:\/\/agileback.lndo.site\/api\/entity_view_mode\/entity_view_mode",
    "entity_form_display--entity_form_display": "http:\/\/agileback.lndo.site\/api\/entity_form_display\/entity_form_display",
    "entity_view_display--entity_view_display": "http:\/\/agileback.lndo.site\/api\/entity_view_display\/entity_view_display"
  }
}
```

So maybe disregard my changes earlier? Or better find a way to incorporate both. There is a probably a way to configure endpoints how you want them with JSONAPI so some flexibility would be useful.

tanc:
Yes, seems sound

packages/source-drupal/src/DrupalSource.js
matt-e-king: I like `apiBase` as well - I'll get that in place.

What is the preferred way to do this?  I can't edit the suggestion made by @tanc, but I can provide a new suggestion which forces me to start a "review". Is it easiest for me just to resolve the conversation and push new code? Or @tanc can you edit your own suggestion and then I can just batch these suggestions into a commit?

packages/source-drupal/src/DrupalSource.js
matt-e-king: If we do the mapping that I proposed as a secondary option, then this change won't be necessary. That said, the mapping I proposed is assuming people will more commonly be using the default JSON:API schema.

packages/source-drupal/src/entities/Entity.js
matt-e-king: Oh great! Yeah, I like that. And yes, if it equals the entity type we can skip the name. So if I'm understanding correct, another example would be `file--file`, would just be `DrupalFile`?

Good to know about the helper function - adding the `s` isn't a big deal to me, but does seem like something the people would demand :)

packages/source-drupal/src/entities/TaxonomyTerms.js
matt-e-king: Doh! I hate how Javascript handles `nulls` in destructuring. It gets me every time. I wish it would see it as a `falsey` value and fallback to the default value that is provided. Ugh.

packages/source-drupal/src/entities/Entity.js
hjvedvik: Yes, I think we can just remove the last part of the string if it equals the first part for all entity types :)

hjvedvik:
> ```
> {
>   "data": [],
>   "links": {
>     "self": "http:\/\/gridsome.lndo.site\/api",
>     "blocks": "http:\/\/gridsome.lndo.site\/api\/blocks"
>     ...
>   }
> }
> ```

@tanc Which version of JSON:API gave this schema?

tanc:
That earlier one was from Contenta (Drupal CMS distribution) with the default content modules enabled. Once starting from scratch (still using Contenta) with no defaults the second schema is the one shown.

In both instances the version of the module was the same, not sure about schema versions if that is a thing

packages/source-drupal/src/DrupalSource.js
```diff
@@ -0,0 +1,201 @@
...
+const Users = require('./entities/Users')
+
+
+class DrupalSource {
+  // defaultOptions merged with this.options in App.vue
+  static defaultOptions() {
+    return {
+      baseUrl: '',
+      views: [], // deprecated
+      entities: [],
```
hjvedvik:
```suggestion
      entities: ['node', 'taxonomy_term', 'file', 'user'],
```

I think we should allways fetch these types as default. And maybe the option should only add more types to the array.

matt-e-king: @hjvedvik the `defaultsDeep` in App.js that merges the `defaultOptions` with the user options doesn't do a concat of the Array. It replaces the `indexes`.

Does this need to be adjust in App.js first?

packages/source-drupal/src/DrupalSource.js
matt-e-king: https://github.com/gridsome/gridsome/pull/41/commits/f84f4ccef46f15555bca9a85a135aa81eb4ab198

packages/source-drupal/src/DrupalSource.js
hjvedvik: I see.. The option should probably be empty if we add them to the default types. You can merge the arrays inside `processEntities` instead.

matt-e-king: Right. Good call.

matt-e-king:
@hjvedvik if I want `source-drupal` to use a dependency already declared in `gridsome/package.json` should I still add it to the `packages/source-drupal/package.json`? 

- **If I leave it out**, then the dependency versions will stay in sync, but it feels too implicit to rely on a dependency declared at the top level. 
- **If I leave it in**, and the versions get out of sync, it appears it will write both versions to `yarn.lock`

hjvedvik:
We should also have an option to set custom routes for each type. Other source plugins does it by having a `routes` option. It could be something like this:

```js
{
  use: '@gridsome/source-drupal',
  options: {
    routes: {
      'node--node': '/blog/:year/:month/:day/:slug',
      'node--custom_type': '/custom/:slug'
    }
  }
}
```

Available route params is currently a bit limited, but I'm working on it in another PR.

matt-e-king:
@hjvedvik I totally agree with custom routes - it's on the roadmap I mentioned in the initial comment.

packages/source-drupal/src/DrupalSource.js
matt-e-king: https://github.com/gridsome/gridsome/pull/41/commits/470568165e6178aaa4bfbbfd62bfdd6804e6476b

tanc:
I was just about to ask about custom routes as this was something I bumped into but I see you're looking into it.

matt-e-king:
@tanc @hjvedvik If we want to support custom keys in the response of the JSON:API schema ([see @tanc's comment](https://github.com/gridsome/gridsome/pull/41#issuecomment-442758818)), should we just combine "[entity mapping](https://github.com/gridsome/gridsome/pull/41#issuecomment-442851384)" and "[custom routes](https://github.com/gridsome/gridsome/pull/41#issuecomment-442915208)" into the same config options, like this:

```
{
  use: '@gridsome/source-drupal',
  options: {
   baseUrl: '',
   entities: [], // add additional entity types
   entityOptions: {
    // default json api schema keys
    'node--article': {
      route: ''
     },
     // custom json api schema keys
     article: {
      route: '',
       entity: 'node'
      }
    }
  }
}
```

This might scale well for customizing entity behavior in the future - but I don't know if this goes against any attempt to set standards for Gridsome plugin `options`.

The alternative would be two different `options`, something like this:
```
{
  use: '@gridsome/source-drupal',
  options: {
    baseUrl: 'http://dev-cctd8.pantheonsite.io',
    entities: [], // add additional entity types
    routes: {
      'node--node': '/blog/:year/:month/:day/:slug',
      'node--custom_type': '/custom/:slug'
    },
    entityMapping: {
      article: 'node',
      files: 'file',
      tags: 'taxonomy_term',
      ...
    }
  }
}
```

tanc:
To me the first option seems sensible and neater but as you say there may be a standard for declaring these things.

hjvedvik:
I've tested the [JSON:API Extras](https://www.drupal.org/project/jsonapi_extras) module (8.x-3.0-rc2), and you can change the entity type for each individual route. Eg. `node--article` to `articles` and the resource path from `/jsonapi/node/article` to `/jsonapi/articles`. You can also create new names for each property on the entities, `body` can be renamed to `content` etc.

So I think the mapping is just solving a small part of the problem. We must either support all possible changes in the schema or just the default settings. One way to support all possible changes is to loop over all the endpoints, fetch the data and simply put the attributes into the store as is no matter which entity type it is. We can still set up relations correctly without any custom mappings because the custom names are reflected everywhere in the schema. But then we should have an `exclude` option to skip certain endpoints. We can also exclude default endpoints for administrative resources by default.

The options could be like this:

```js
{
  use: '@gridsome/source-drupal',
  options: {
    routes: {
      'node--article': '/blog/:slug', // uses default schema settings
      'product': '/shop/:slug' // renamed 'node--product' to 'product'
    },
    exclude: [
      'blocks', // 'block--block' was renamed to 'blocks', so exclude it manually
      'my_entity--*' // exclude all custom entities manually
    ]
  }
}
```

What do you think?

matt-e-king:
@hjvedvik This is sounding good. Originally I was looping over EVERYTHING, but I was getting so many 403/405 because you have to grant special permissions for a lot of those URLs so I felt the developer would get frustrated (I guess this is fine as long as 403/405 network errors don't stop the build from running). 

But, like you mentioned, if we can come up with a good list of default excludes (like the admin resources) that we know won't we useful - and then provide the user an option to add more excludes, that should hopefully resolve that. And any entity class we haven't coded yet (/src/entities/*) can just be in the default excluded list until it is supported.

**Fields/Attributes:**
Regarding the ability to change even the fields, like `body` to `content`, I think when we destructure the `attributes` we can check for all those defaults first - `title, content, date, excerpt, slug` - and then have some intelligent fallbacks.

**One question:**
To make things easier, currently we using the whole `key` from the json api schema object to name the GraphQL Type:

```
node--article --> NodeArticle  // default node naming convention
product --> Product  // custom node naming
taxonomy_term--tags --> TaxonomyTermTags  // default term naming
category --> Category  // custom term naming
```
In one iteration that I did I would `split` the `key` and use everything after the `--` for the GraphQL Type. But my worry was that a user would name two entity types the same name (`node---article, taxonomy_term--article`). 

So in light of this conversation, how should we handle this? Should we still just use the `key` from the JSON:API schema and put the responsibility on the use since we are supporting custom naming in JSON:API extras?

Thanks, guys!

hjvedvik:
I see.. The plugin could have an `auth` option for authenticated requests. And yes, just skipping requests which responds with `403/405` should be okay :)

We might remove `title`, `date`, `content` and `excerpt` from the GraphQL schema sometime and just rely on custom fields instead because they are a little too specific and doesn't work for all sources. So all attributes could go directly into the store as custom fields. But the store is depending on the `date` now, so we must have some fallbacks.

And I think we can use the names as they are in the schema, like in your examples. We can consider using only the entity name if it equals the type (`file--file -> File`). I think it also would be good to prefix the names with `Drupal` or a custom name from a `typeName` option. They will then become `DrupalNodeArticle`, `DrupalFile` etc.

tanc:
Another couple of things to throw in the mix: paragraph entries and media entities

Media entities are Drupal core's way of handling things like images as well as remote or embeddable media (like YouTube videos).

Paragraphs are a common way of having fieldable entities within a node and is useful for structured data.

When I created a quick paragraphs import file based off the node one I noticed that fields in a paragraph weren't necessarily prefixed with field_ and therefore weren't added as custom fields to graphql. Once we have a firmer set of import files I'm happy to take a look at paragraphs

matt-e-king:
Awesome, thanks guys! I'm hoping to drum up some time over the next couple of days to work through these remaining items.

And yes, thanks @tanc, I don't claim to be a Drupal expert by any account, so I look forward to your contributions for the more detailed, nuanced Drupal-ish stuff :)

matt-e-king:
@hjvedvik I'm hoping you can save me some troubleshooting... it seems like the GraphQL server is choking when I attempt to create nodes with deeply nested complex types.

One level deep works just fine:
```
{
  id: '12345',
  body: {
    value: 'foo',
    summary: 'bar'   
  } 
}
```
I can traversed `body` no problems in the graphql explorer. I don't know the exact combination, but it seems if `body` had a nested `object` or `array`, it fails.

I don't actually get an error, everything appears to run, but this is what I see in the explorer:
screen shot 2018-11-30 at 4 59 08 pm

Any thoughts?

(FYI - super close to having everything we've discussed in place!)

hjvedvik:
I can check this out tomorrow :) Gridsome core should be able to transform all custom fields into a GraphQL schema, also nested objects and arrays. Is the API endpoint you're testing open somewhere?

matt-e-king:
> Is the API endpoint you're testing open somewhere?

Yes, `tour--tour` and `view--view` I know for sure break for me within my example project:

http://dev-cctd8.pantheonsite.io/jsonapi/view/view
http://dev-cctd8.pantheonsite.io/jsonapi/tour/tour

They both required setting up the permissions correctly in Drupal.

I'm planning on pushing my changes shortly - so that might be easier to test with and you can just use my `baseUrl` option of `http://dev-cctd8.pantheonsite.io`.

hjvedvik:
I think the GraphQL server crashes because some of the keys in the schema contain dashes, like `views-main`. I can make Gridsome core converts all keys into valid schema field names. Gridsome currently camelCases root level fields but should convert all nested fields too.

matt-e-king:
Updates:

- Added supported for custom routes in `gridsome.config.js`
- Dropped support for "supported entities" and instead added support for "excluded entities"
-- see `constant.js` for the default excludes
-- @tanc can you take a look at `constants.js` and provide some insight on which entities will never be much use to the end user?
- Added default `typeName` with value of "Drupal"
- Removed redundant entity name; `file--file` just becomes `DrupalFile`, not `DrupalFileFile`
- Removed the need for individual entity classes; all instance methods are now in `Entity.js`
-- That said, it will be super easy to add individual entity class overrides in the future if need **we could even open this up to the user if they want to provide a "render" class for a specific entity type**

Here is an example `gridsome.options.js`:
```
module.exports = {
  plugins: [
    {
      use: '@gridsome/source-drupal',
      options: {
        baseUrl: 'http://dev-cctd8.pantheonsite.io',
        entities: [],
        excludes: ['menu--menu'],
        routes: {
          'node--article': '/articles/:slug',
          'taxonomy_term--tags': '/tags/:slug'
        }
      }
    }
  ]
}
```

**I haven't added the ability to exclude as a wildcard yet - like node--***

tanc:
Excellent work again @matt-e-king 👍 

I'm testing this out this morning and all the entity stuff looks good. My list of additional excludes is:
```
'menu--menu',
'block--block',
'entity_browser--entity_browser',
'jsonapi_resource_config--jsonapi_resource_config',
'configurable_language--configurable_language',
'language_content_settings--language_content_settings',
'media_type--media_type',
'paragraphs_type--paragraphs_type',
'responsive_image_style--responsive_image_style',
'oauth2_token_type--oauth2_token_type',
'action--action',
'taxonomy_vocabulary--taxonomy_vocabulary',
'user--user',
'menu_link_content--menu_link_content',
'base_field_override--base_field_override',
```

Of these the following might be useful to exclude by default:
```
'base_field_override--base_field_override',
'action--action',
'oauth2_token_type--oauth2_token_type',
'entity_browser--entity_browser',
'jsonapi_resource_config--jsonapi_resource_config',
'configurable_language--configurable_language',
'language_content_settings--language_content_settings',
```

Can a user 'un-exclude' or re-include entities excluded by default?

tanc:
I'm also having some trouble with getting routes to work, but it may well be my lack of understanding of the Template system that is at fault. I have a `DrupalNodeBlogPosts` (why is it pluralised?) GraphQL node which I can query fine using the GraphQL explorer. For example:
```
{
  drupalNodeBlogPosts (path: "/node-blog-posts/test-blog-post") {
    title
    slug
    path
  }
}
```
gets me:
```
{
  "data": {
    "drupalNodeBlogPosts": {
      "title": "Test blog post",
      "slug": "test-blog-post",
      "path": "/node-blog-posts/test-blog-post"
    }
  }
}
```

In my `gridsome.config.js` I've got the routes set up like:
```
routes: {
  'node--blog_post': '/:slug'
}
```
and confirmed with some console logging that the route is picked up when pulling in blog posts.

I've created a template file `src/templates/DrupalNodeBlogPosts.vue` with the following simple content:
```
<template>
  <Layout :title="$page.post.title">
    <h3>{{ $page.post.title }}</h3>
  </Layout>
</template>

<page-query>
query Post ($path: String!) {
  post: drupalNodeBlogPosts (path: $path) {
    title
  }
}
</page-query>

<script>
export default {
  metaInfo () {
    return {
      title: this.$page.post.title
    }
  }
}
</script>
```

Whatever path I hit I can't seem to get this template file to work. Looking back at the result of the GraphQL query the path for my sample blog post is `/node-blog-posts/test-blog-post`. But in theory, shouldn't it be `/test-blog-post` if I've specified that route for it in `gridsome.config.js`? Either way, whatever path I hit I can't seem to get that template to trigger and always get a 404. Any thoughts?

tanc:
@hjvedvik maybe the above is more a question for you? I'm still confused about how `path` and `route` interact. In the docs for `addNode` it says for `path`: Optional path to use when not having a dynamic route.

Yet in the dynamic template the path is used to filter for the node. The problem is the path is not the route so they can't match. Or am I missing something?

tanc:
Odd... things _are_ hooking up correctly now, the path attribute is correct and based off the slug now. I'm not sure what I've changed but there must have been a mismatch somewhere in my code.

I'll keep testing!

matt-e-king:
Awesome, @tanc - thanks for keeping us in the loop.

> Can a user 'un-exclude' or re-include entities excluded by default?

@tanc Maybe the excludes shouldn't work as a merge, maybe it should be an override? So if the user decides to add their own excludes, then none of the defaults would exists and they are responsible for all excludes. 

Plus, if someone uses JSON:API extras and renames their endpoints, none of the default excludes would work anyways.

tanc:
Yes I think override is more flexible as some people may need certain entities.

Is there a way to get the entity type in the GraphQL query? I'm trying to think of a way to iterate over a set of paragraph entities that are children of the parent node entity and use a different component for each instance depending on the name of the entity.

For example, something like this:
```
<template>
  <Layout :title="$page.post.title">
    <h3>{{ $page.post.title }}</h3>
    <component
      v-for="contentSection in $page.post.contentSections"
      :is="contentSection.name"
      :key="contentSection.id"/>
  </Layout>
</template>

<page-query>
query Post ($path: String!) {
  post: drupalNodeBlogPosts (path: $path) {
    title
    contentSections {
      name
      id
    }
  }
}
</page-query>
```
The above doesn't work as `name` is not a know property of the GraphQL node. The `contentSections` is the name of the Drupal field which references Paragraph entities containing the bulk of the node's content.

matt-e-king:
@tanc I can update this so the `type` is added as a field on each node.

It would be something like `node--blog_post` or just `blog_post` depending on how you have the JSON:API endpoint setup. 

Is that what you're hoping for?

tanc:
@matt-e-king that would be perfect.

At the moment there is a `type` on the node but this is UUID which matches up with the corresponding endpoint in the API. For example, on a Drupal node content type

```
{
  drupalNodeBlogPosts (id: "1") {
    title
    type
  }
}
```
Gives:
```
{
  "data": {
    "drupalNodeBlogPosts": {
      "title": "Test blog post",
      "type": [
        "6d1bab31-e9cf-40df-8d98-83fc846bf25c"
      ]
    }
  }
}
```
In the API endpoint: `api/node_type/node_type` the matching `id` for that UUID looks like:
```
{
  "type": "node_type--node_type",
  "id": "6d1bab31-e9cf-40df-8d98-83fc846bf25c",
  "attributes": {
    "uuid": "6d1bab31-e9cf-40df-8d98-83fc846bf25c",
    "langcode": "en",
    "status": true,
    "dependencies": {
    "module": [
      "menu_ui"
    ]
  },
  "third_party_settings": {
    "menu_ui": {
    "available_menus": [],
    "parent": ""
    }
  },
  "name": "Blog post",
  "type": "blog_post",
  "description": "A blog post is a piece of writing or other item of content displayed in a blog.",
  "help": "",
  "new_revision": true,
  "preview_mode": 0,
  "display_submitted": false
},
"links": {
  "self": "http:\/\/agileback.lndo.site\/api\/node_type\/node_type\/6d1bab31-e9cf-40df-8d98-83fc846bf25c"
  }
}
```

Is it worth connecting these together via the ID so that the extra information about that content type is available? Or not worth the hassle?

matt-e-king:
@tanc actually, `node_type` is listed as a `relationship` in the response and this plugin is written to create GraphQL Connections out of ALL the relationships defined in each response (as long as the GraphQL content type exists).

So the fix was pretty easy for me since I am not renaming the JSON:API endpoints, I just had to remove `node_type--node_type` from the default excludes and then I could run the following GraphQL query:
```
query Articles {
  allDrupalNodeArticles (perPage:100) {
    edges {
      node {
        title,
        path,
        fieldImage {
          filename
        },
        fieldTags {
          id
        },
        type {
          type
        }
      }
    }
  }
}
```

Notice:
```
type {
  type
}
```
This should work for you now with the most recent commit I pushed.

tanc:
Oh nice! That works well for Drupal Nodes. Unfortunately Paragraphs don't get picked up, the logs say:
```
DrupalParagraphsTypes has no values in response
DrupalParagraphsTypes initialzed...
```

This an example of the output from the `/api/paragraphs_type/paragraphs_type` endpoint in case that helps:
```
{
  "type": "paragraphs_type--paragraphs_type",
  "id": "a586d82e-ab69-4de4-a769-feec5bb87539",
  "attributes": {
    "uuid": "a586d82e-ab69-4de4-a769-feec5bb87539",
    "langcode": "en",
    "status": true,
    "dependencies": [],
    "id": "text_block",
    "label": "Text block",
    "icon_uuid": null,
    "description": "",
    "behavior_plugins": []
  },
  "links": {
  "self": "http:\/\/agileback.lndo.site\/api\/paragraphs_type\/paragraphs_type\/a586d82e-ab69-4de4-a769-feec5bb87539"
  }
},
```

tanc:
Below is an example paragraph from the endpoint. The relationships should tie the paragraph types to the paragraph if the import isn't skipped.
```
{
  "type": "paragraph--text_block",
  "id": "133c2ee8-6f44-46ef-8541-6b12c9cf22d9",
  "attributes": {
    "id": 1,
    "uuid": "133c2ee8-6f44-46ef-8541-6b12c9cf22d9",
    "revision_id": 3,
    "langcode": "en",
    "status": true,
    "created": 1543497695,
    "parent_id": "2",
    "parent_type": "node",
    "parent_field_name": "content_sections",
    "behavior_settings": "a:0:{}",
    "default_langcode": true,
    "revision_translation_affected": true,
    "field_text": {
      "value": "This is a test",
      "format": "restricted_html",
      "processed": "<p>This is a test</p>\n"
    }
  },
  "relationships": {
    "type": {
      "data": {
        "type": "paragraphs_type--paragraphs_type",
        "id": "a586d82e-ab69-4de4-a769-feec5bb87539"
      },
      "links": {
        "self": "http://agileback.lndo.site/api/paragraph/text_block/133c2ee8-6f44-46ef-8541-6b12c9cf22d9/relationships/type",
        "related": "http://agileback.lndo.site/api/paragraph/text_block/133c2ee8-6f44-46ef-8541-6b12c9cf22d9/type"
      }
    },
    "uid": {
      "data": {
        "type": "user--user",
        "id": "611c5ff8-c5dd-451c-89cb-2fe171a75487"
      },
      "links": {
        "self": "http://agileback.lndo.site/api/paragraph/text_block/133c2ee8-6f44-46ef-8541-6b12c9cf22d9/relationships/uid",
        "related": "http://agileback.lndo.site/api/paragraph/text_block/133c2ee8-6f44-46ef-8541-6b12c9cf22d9/uid"
      }
    },
    "revision_uid": {
      "data": {
        "type": "user--user",
        "id": "611c5ff8-c5dd-451c-89cb-2fe171a75487"
      },
      "links": {
        "self": "http://agileback.lndo.site/api/paragraph/text_block/133c2ee8-6f44-46ef-8541-6b12c9cf22d9/relationships/revision_uid",
        "related": "http://agileback.lndo.site/api/paragraph/text_block/133c2ee8-6f44-46ef-8541-6b12c9cf22d9/revision_uid"
      }
    }
  },
  "links": {
    "self": "http://agileback.lndo.site/api/paragraph/text_block/133c2ee8-6f44-46ef-8541-6b12c9cf22d9"
  }
},
```

hjvedvik:
Good work @matt-e-king :) 

If the `exclude` option will override the default excluded endpoints, then we could move the plugin code over to a `gridsome.server.js` file, and then export the defaults from an `index.js` file. That will allow devs to do something like this if they want to keep the default ones:

```js
const { exclude } = require('@gridsome/source-drupal')

module.exports = {
  plugins: [
    {
      use: '@gridsome/source-drupal',
      options: {
        exclude: [...excludes, 'exclude--me']
      }
    }
  ]
}
```

I'm working on a better way to create references in another [PR](/gridsome/gridsome/pull/50), which I'll merge in within a few days. I think it will make it much easier for you to set up the references from JSON:API.

hjvedvik:
@matt-e-king @tanc Just created a channel for Drupal over at the Gridsome Discord app, if you want to discuss there instead :)

matt-e-king:
> If the `exclude` option will override the default excluded endpoints, then we could move the plugin code over to a `gridsome.server.js` file

I'll give this a shot.

matt-e-king:
@hjvedvik I just pushed updates for https://github.com/gridsome/gridsome/pull/41#issuecomment-443863562.

Also, I'll keep an eye on https://github.com/gridsome/gridsome/pull/50, looking forward to that!

Is it too soon to start drafting a new README? And let me know if you want help with a `gridsome-starter-drupal`.

segovia94:
You may have already done this, but you might want to check out how Gatsby integrated their source plugin for Drupal https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-drupal. It assumes JSON:API is enabled on the Drupal site. I know a lot in the Drupal community are using this.

Also Contenta has a public API for demo stuff at https://back-end.contentacms.io so you can query articles for example at https://back-end.contentacms.io/api/articles. Might be helpful to some when evaluating.

matt-e-king:
@segovia94 Yes, this integration is built with JSON:API as a required dependency.

matt-e-king:
@hjvedvik will do! Thanks.

matt-e-king:
All should be good

matt-e-king:
@hjvedvik Happy New Year!

I hate to even ask this, but I realized this morning my `git config user.email` was set to the wrong email address for this project. How opposed are you to running [this](https://help.github.com/articles/changing-author-info/) against my working branch to correct it to my github email address?

I understand the risks of this, just figured I'd ask :)

hjvedvik:
Happy New Year to you too @matt-e-king! :)

Maybe it's not necessary to change the previous commits. The commits in this PR will be squashed into one commit when it's merged. And I think that commit will belong to the user who created the PR. But I'll investigate it a bit further.

hjvedvik:
@matt-e-king Can you rename the `src` folder to `lib`? Everything else look good and I think it can be merged and published pretty soon :)

PR #46, opened 2018-11-17
feat: make style loader options configurable

maoberlehner:
I need the possibility to configure the SCSS loader options in all of my projects. I guess it is possible via chain-webpack but I couldn't get it to work. I think exposing the loader options directly makes this a lot easier.

### DIFF
--- a/gridsome/lib/webpack/createBaseConfig.js
+++ b/gridsome/lib/webpack/createBaseConfig.js
@@ -273,6 +273,7 @@ module.exports = (app, { isProd, isServer }) => {
   }
 
   function createCSSRule (config, lang, test, loader = null, options = {}) {
+    const { css = {}, postcss = {}} = projectConfig.css.loaderOptions
     const baseRule = config.module.rule(lang).test(test)
     const modulesRule = baseRule.oneOf('modules').resourceQuery(/module/)
     const normalRule = baseRule.oneOf('normal')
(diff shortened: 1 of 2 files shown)


maoberlehner:
Also updated documentation: https://github.com/gridsome/gridsome.org/pull/31

hjvedvik:
Thanks :) I waited with implementing this because I was unsure how the config should look like. Would it be a good idea to have it more similar to the [Vue CLI CSS config](https://cli.vuejs.org/config/#css-loaderoptions)? I like that they have CSS specific config grouped in a `css` property and the loader options in `css.loaderOptions`.

maoberlehner:
I think making it similar to the Vue CLI config would be great! I'll see if a can update this pull request accordingly.

maoberlehner:
I've updated the pull request!

maoberlehner:
Yes, like this though:

```js
module.exports = {
  plugins: [],
  css: {
    loaderOptions: {
      scss: {
        data: `@import '${path.resolve(__dirname, 'src', 'assets', 'scss').replace(/\\/g, '/')}/_global.scss';`
      }
    }
  }
};
```

See: https://deploy-preview-31--gridsome-org-live.netlify.com/docs/config#cssloaderoptions

lusarz:
Thank you, I like it

gridsome/lib/webpack/createBaseConfig.js
```diff
@@ -126,14 +126,12 @@ module.exports = (app, { isProd, isServer }) => {
...
 
-  createCSSRule(config, 'css', /\.css$/)
-  createCSSRule(config, 'postcss', /\.p(ost)?css$/)
-  createCSSRule(config, 'scss', /\.scss$/, 'sass-loader', projectConfig.scss)
-  createCSSRule(config, 'sass', /\.sass$/, 'sass-loader', Object.assign({ indentedSyntax: true }, projectConfig.sass))
-  createCSSRule(config, 'less', /\.less$/, 'less-loader', projectConfig.less)
-  createCSSRule(config, 'stylus', /\.styl(us)?$/, 'stylus-loader', Object.assign({
-    preferPathResolver: 'webpack'
-  }, projectConfig.stylus))
+  createCSSRule(config, 'css', /\.css$/, null, projectConfig.css.loaderConfig.css)
```
lusarz: I've got error on this line - cannot read `css` of undefined when used [this config](https://github.com/gridsome/gridsome/pull/46#issuecomment-450556456). Seems like there is some confusion between `loaderConfig` vs `loaderOptions` names.

maoberlehner: I'll take a look!

Issue #47, opened 2018-11-17
TypeError: Cannot read property 'addReference' of undefined


andrewhl:
I'm using Gridsome ^0.3.0, and `@gridsome/source-wordpress`, ^0.1.0, and I'm running into the following error upon running `gridsome develop`:

```bash
TypeError: Cannot read property 'addReference' of undefined
    at WordPressSource.fetchWordPressContent (/Users/andrew/code/SpiralPraxis/marketing/node_modules/@gridsome/source-wordpress/index.js:99:20)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

I've got a vanilla installation. I haven't modified any files except `gridsome.config.js`, and that only to provide a URL to my wordpress site, which is enabled to serve wp-json endpoints.

Here's my config file:

```javascript
module.exports = {
  plugins: [{
    use: '@gridsome/source-wordpress',
    options: {
      baseUrl: 'https://mywordpresssite.com', // required
      typeName: 'WordPress', // GraphQL schema name (Optional)
      perPage: 100, // How many posts to load from server per request (Optional)
      concurrent: 10, // How many requests to run simultaneously (Optional)
      routes: {
        post: '/:year/:month/:day/:slug', //adds route for "post" post type (Optional)
        post_tag: '/tag/:slug' // adds route for "post_tag" post type (Optional)
      }
    }
  }]
}
```

Here's the relevant code from `@gridsome/source-wordpress/index.js`:

```javascript

    for (const type in taxonomies) {
      const options = taxonomies[type]
      const typeName = makeTypeName(type)
      const route = routes[type] || `/${options.rest_base}/:slug`

      restBases.taxonomies[type] = options.rest_base

      for (const type of options.types) {
        const postTypeName = makeTypeName(type)
        const collection = getContentType(postTypeName)

        collection.addReference(options.rest_base, { typeName, key: '_id' })
      }

      addContentType({ typeName, route })
    }

```

Not sure what to do. I can't see anything in the documentation that clarifies this issue. Help would be appreciated.

hjvedvik:
Your PR is merged and published :) Can you verify that it is working now?

PR #49, opened 2018-11-18
Fix "Cannot read property 'addReference' of undefined" error

andrewhl:
See [this issue](https://github.com/gridsome/gridsome/issues/47) for more details. Adding a guard for line 97's collection seems to fix the problem.


hjvedvik:
Thanks :)

PR #64, opened 2018-11-30
collection.addSchemaField fix

matt-e-king:
@hjvedvik According to the documentation for `collection.addSchemaField`, the callback function should receive an object with a property of `graphql`. 

This wasn't working for me - do these modifications to the `createType.js` fix that?


hjvedvik:
Yes, that should fix it. I wrote the docs before I implemented it, so must have forgotten that one :)

PR #65, opened 2018-12-04
Allow router-link classes to be adjusted

0aveRyan:
This helps with frameworks like Bulma, Bootstrap or Pure with predefined active classes.


hjvedvik:
Yes, would be nice to customize the active classes :) Can you set the class names as default prop values instead? Like this:

```js
activeClass: { type: String, default: 'active' },
exactActiveClass: { type: String, default: 'active--exact' }
```

0aveRyan:
@hjvedvik fixed!

PR #77, opened 2018-12-18
Current node fields as query context

hjvedvik:
This PR makes it possible to use custom fields as query variables in templates.


startupengine:
Is there any documentation for this?

PR #84, opened 2018-12-20
GraphQL filter argument

### LINKED ISSUE #56: GraphQL query filter not working
I am trying to `filter` a list of blog posts by `author` and `tag`, but GraphQL is throwing error `Unknown argument \"filter\" on field \"allGhostBlogPost\" of type \"Query\"."`. I am current running gridsome version `0.3.4`.

```
query AuthorIndex ($slug: String!) {
  allGhostBlogPost (filter: {primaryAuthor: {slug: {eq: $slug}}}) {
    edges {
      node {
        id
        title
        path
        content
        excerpt
        date
        primaryTag {
          url
          name
          slug
        }
        primaryAuthor{
          name
          url
          slug
          profile_image
        }
      }
    }
  }
}
```

I understand that  `filter` argument is not included in default schema. So I assume I can add it in some way.

Here are a few references from Stackoverflow,

- [add a custom field and corresponding resolver](https://stackoverflow.com/questions/48352814/graphql-query-some-record-error-unknown-argument)
- [you would need to change the Query type definition to accept filter string](https://stackoverflow.com/questions/48340191/which-syntax-for-filtering-queries-in-graphiql)

hjvedvik:
This PR implements a `filter` argument which will include most of the [LokiJS find operators](https://github.com/techfort/LokiJS/wiki/Query-Examples).

#### Example query

```graphql
query FeaturedProducts {
  allProduct (
    filter: {
      tags: { contains: ["featured"] }
      price: { between: [49, 99] }
    }
  ) {
    edges {
      node {
        title
        price
      }
    }
  }
}
```

Closes #56 

Todo: 
- [x] Filter `belongsTo` by `typeName`
- [x] Calculate pagination for all collections and the `belongsTo` field
- [x] Documentation


arttemiuss43:
Is there any way to use Vue properties as filter values?

hjvedvik:
No, only fields from the current node will be available as query variables because the query result is rendered as a static JSON file. But you can filter the data further in your Vue components :)

arttemiuss43:
Thank you for the answer.
But if I will filter data further how can I transform it to the graphql schema? Do I have to override the $page property and manually create edges and so on? Or is there more natural way? If I'm fundamentally wrong could you please tell a basic workflow of doing data filtering and sorting?

PR #92, opened 2019-01-02
Storyblok CMS as source.

TinyChief:
Hello!

Pull requesting for the first time!))
Is it okey or changes are needed?


hjvedvik:
Thank you! The PR looks very good, but I'll try to take a closer look and test it this weekend :)

TinyChief:
I didn't fully understand how you manage plugins, I mean installing, registering as npm modules and so on. Looks like I added `storyblok-js-client` module to dependencies of gridsome, so that is not right. Shall it be there or not?

hjvedvik:
That is correct :) Lerna creates only a single lock file for all the packages in this repo.

hjvedvik:
Hi, sorry for the delay on this. I've been testing the plugin an I think it will work pretty good :) Just got one question. I tried to create a blog, but the overview page is mixed with the articles. Should we skip items which are `is_startpage: true`? Or do something else with them?

TinyChief:
This looks a bit cleaner. Also you can query data any possible way. What do you think?

hjvedvik:
I think it looks better :) I have tested a bit myself and did some changes to also support data, etc. Is it okay if I update this PR with my changes? Would be great if you're able to test it further.

TinyChief:
That's ok!

I'm using Storyblok as headless CMS at my project, so, yeah, I will test it))

hjvedvik:
Updated the PR now. Would be great if you are able to test it :) I've updated the readme with the new options.

altryne:
👋 Interested in testing this out, any progress/update on this or was this effort abandoned?

hjvedvik:
@altryne The code in this PR is probably a bit old, But the Storyblok team has made their own official source plugin that you can try out: https://gridsome.org/plugins/gridsome-source-storyblok

I'll close this for now :)

Issue #98, opened 2019-01-04
source-contentful does not register refs for single Link of type Entry


crsten:
## Description

The source-contentful plugin does noe register refs if it is a single reference. The plugin only checks for the following:
```js
if (field.items && field.items.type === 'Link' && field.items.linkType === 'Entry')
```

It should also check this:
```js
if(field.type === 'Link' && field.linkType === 'Entry')
```

### Steps to reproduce

1. Create ContentType in Contentful with single reference
2. Check the output in GraphQL explorer 
3. You should see that the single reference did not resolve correctly

### Expected result

It should resolve the reference in graphql

### Actual result

It resolves as an array containing the id

**PR is on the way**

crsten:
will be fixed in #102

PR #99, opened 2019-01-04
fix(source-contentful): Add single references from contentful to "refs"

crsten:
Added an extra check for single references and added it to refs in order for it to resolve correctly

#98


crsten:
This PR is obsolete and will be fixed with #102

PR #100, opened 2019-01-04
feat: Add option to add tagline to homepage title [wip]

tomtev:
The default homepage title now is `Sitename - Sitename`.
It should be `Tagline - Sitename`, and if no tagline is set just `Sitename`

Todo:

- [ ] Make sure tagline is only used on homepage
- [ ] Fallback to `Sitename` only if tagline or custom vue meta title for homepage is not set.


m1guelpf:
Any updates on this?

PR #101, opened 2019-01-04
feat: Add a simple example plugin

tomtev:
A simple example plugin showing how to add data to data store.

Todo: 
- [ ]  Show how to use client api


tomtev:
This will be added to another repo

PR #102, opened 2019-01-04
Contentful: Refactor and update with new reference API

hjvedvik:


crsten:
Did also test the following without errors:

- Multiple Entries of different type (Reference/GraphQLUnion)

Looks like a solid PR

nolfranklin:
Exploring with potential issue:

```
"@gridsome/source-contentful": "^0.2.1",
"gridsome": "^0.4.0"
```

I have a Contentful "References, many" field named "blocks" in my "Page" content type that references a few other content types (text, jumbotron, etc.) When I query I can't seem to get the multiple types to show up. Only a single reference type "text" fields show up in the block { fields {} } query but interestingly enough returns the correct number of referenced items. 

Using __typename also returns "ContentfulPageBlocks" instead of the reference content types.

Appreciate the insights! :)

PR #115, opened 2019-01-10
feat(routeKeys): adding support for nested routeKeys via '__'

### LINKED ISSUE #114: Add support for nested field values in route template
## Summary

It would be very handy if it was possible to use deep properties inside the route templates.

### Basic example

Object from source: 
```
{
  title: 'Welcome',
  metadata: {
    slug: '/welcome-sample'
  }
}
```

Route template:

`/blog/:metadata__slug`

### Motivation

It would make routing setup more flexible. In my case i use Contentful as source where i have a type that contains the slug, which is referenced into Pages below ex. `metadata`. That way it is also easier to keep the slugs unique since contentful has an option "unique" that forces all entries of the same type to contain only one occurence.

crsten:
closes #114


tomtev:
Thank you @crsten 🌟

Issue #58, opened 2018-11-27
Use within Docker containers


tanc:
I'm experimenting with running Gridsome within a set of docker containers. The first issue is the localhost is the default host which stops the browser from connecting with the server within the containers. This is resolvable by specifying 0.0.0.0 as the host using `gridsome develop -h 0.0.0.0`

The next problem is the served application can't connect with the GraphQL server, the browser throws the following:
```
fetch.js?1477:12 OPTIONS http://0.0.0.0:8080/___graphql net::ERR_CONNECTION_REFUSED
```

If I specify an actual hostname (which resolves to the IP of the docker container) then I get errors running the server with `gridsome develop -h gridsome.lndo.site`:

```
Gridsome v0.3.4

Initialize - 0.06s
Load sources - 0.1s
Create GraphQL schema - 0.04s
Generate code - 0.02s
Bootstrap finish - 0.23s
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRNOTAVAIL 172.24.0.4:9000
    at Server.setupListenHandle [as _listen2] (net.js:1343:19)
    at listenInCluster (net.js:1401:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1510:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:72:10)
```

Any ideas how to proceed?

hjvedvik:
Would it help if you could set a custom URL for the GraphQL server in `gridsome.config.js`?

tanc:
Yes, I think that would do it. I'm also seeing connection problems to the server on port 9000 but it hasn't seemed to cause problems so far, I'm not sure what that does?

hjvedvik:
Port `9000` is for sockjs which makes the `page-query` re-run whenever a node changes on the server. It's not possible to set a custom port atm, but there could be an option in `gridsome.config.js` for that too.

tanc:
I think the port is fine but like the graphql server I probably need to be able to specify a host the browser understands

tanc:
Oh just a note, in the docs it says you can specify a host in the config but this didn't work for me. The only way a new host got picked up was to specify it with the -h switch on the commandline.

tanc:
> Would it help if you could set a custom URL for the GraphQL server in `gridsome.config.js`?

Any luck implementing this?

carmanchris31:
[...]

PR #123, opened 2019-01-12
feat(webpack): make it possible to add enviroment variables

### LINKED ISSUE #88: Environment variables
We should be able to specify environment specific variables that will be available in Vue components? May be something like `development.env` and `production.env` in the root of a project?

isoppp:
fix #88 and #58 

Code and specification referenced Nuxt.js ([doc](https://nuxtjs.org/api/configuration-env/)).

## Description

- Add `env` config to `gridsome.config.js`
- Make it possible to access these configs in Vue

- Attach environment variables start with `GRIDSOME_ENV` to `process.env`

- Add `process.env.NODE_ENV` and `process.isProduction` to base env config.

---

If need document, I will try to create it after review.

### DIFF
--- a/gridsome/app/page-query/fetch.js
+++ b/gridsome/app/page-query/fetch.js
@@ -9,7 +9,7 @@ export default (route, query) => {
     const variables = { ...route.params, path: route.path }
 
     return new Promise(resolve => {
-      fetch(GRAPHQL_ENDPOINT, {
+      fetch(process.env.GRAPHQL_ENDPOINT, {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify({ query, variables })
--- a/gridsome/app/page-query/sock.js
+++ b/gridsome/app/page-query/sock.js
@@ -1,7 +1,0 @@
-/* global SOCKJS_ENDPOINT */
-
-import SockJS from 'sockjs-client'
-
-const sock = new SockJS(process.env.SOCKJS_ENDPOINT)
-
-export default sock
--- a/gridsome/app/page-query/sock.js
+++ b/gridsome/app/page-query/sock.js
@@ -2,6 +2,6 @@
 
 import SockJS from 'sockjs-client'
 
-const sock = new SockJS(SOCKJS_ENDPOINT)
+const sock = new SockJS(process.env.SOCKJS_ENDPOINT)
 
 export default sock
--- a/gridsome/lib/app/App.js
+++ b/gridsome/lib/app/App.js
@@ -36,7 +36,7 @@ class App {
 
   initProcessEnv () {
     const env = process.env.NODE_ENV || 'development'
-    const envFile = path.join(process.cwd(), `./.env.${env}`)
+    const envFile = this.resolve(`./.env.${env}`)
 
     let parsed = {}
     try {
@@ -47,10 +47,7 @@ class App {
       }
     }
 
-    process.env = {
-      ...process.env,
-      ...parsed
-    }
+    Object.assign(process.env, parsed)
   }
 
   async bootstrap (phase) {
--- a/gridsome/lib/app/App.js
+++ b/gridsome/lib/app/App.js
@@ -1,8 +1,10 @@
 const path = require('path')
+const fs = require('fs')
 const isUrl = require('is-url')
 const Router = require('vue-router')
 const autoBind = require('auto-bind')
 const hirestime = require('hirestime')
+const dotenv = require('dotenv')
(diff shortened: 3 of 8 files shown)


hjvedvik:
Thanks, this will be a nice feature :)

I think we can just let the env prefix be `GRIDSOME_` since they already are environment variables. And what do you think about including the `dotenv` package to support `.env` files too?

The `develop` command adds `SOCKJS_ENDPOINT`, `GRAPHQL_ENDPOINT` and `GRAPHQL_WS_ENDPOINT` as global variables through the webpack define plugin. But it would be better to add them to `process.env` and let people override them somehow. Specially for this use case #58.

isoppp:
Thank you for confirmation :D

I tried to fix those issue.

- Change the name prefix form GRIDSOME_ENV to GRIDSOME_.

- Add dotenv package support.

But perhaps it may confuse some users, because it's difficult to merge `dotenv` and `config.env` when it register webpack.
For example, `SOCKJS_ENDPOINT` etc added this time can be overwritten from `gridsome.config.env`, but from `.env` it can not overwrite.
If you want to avoid this confusion it may be better to detach dotenv as a plugin or need some good idea.

- Make it possible to specify dev endpoints

hjvedvik:
Great work! Yes, it's probably a good idea to just use `dotenv.parse` directly so you have more control. And would it be an idea to parse and merge with `process.env` before `gridsome.config.js` is loaded in `app/loadConfig.js`? That will probably make the `env` config unnecessary since the `GRIDSOME_*` plus the `develop` variables can be added to webpack automatically and `process.env` has all the other variables in Node.js. Or what do you think?

#### Example usage:

```
// .env
GRIDSOME_API_URL=https://api.example.com
API_KEY=1234567890

// gridsome.config.js
module.exports = {
  plugins: [
    {
      use: 'gridsome-source-plugin',
      options: {
        apiKey: process.env.API_KEY
      }
    }
  ]
}

// src/pages/Index.vue
export default (
  async mounted () {
    const res = await axios.get(process.env.GRIDSOME_API_URL)
  }
}
```

isoppp:
wow, nice idea! :D
I fixed it!

And, made it possible switch env file by NODE_ENV.
(.env.development and .env.production)

Please confirm again :)

hjvedvik:
Looks good :) Just a few thoughts; Use `context` instead of `process.cwd()`. We should also fallback to `.env` if no `.env.${env}` file exists. And maybe it is better to use `Object.assign(process.env, parsed)` to keep the original reference to `process.env`.

isoppp:
Thank you for confirmation! I fixed those issues :)

isoppp:
above commit resolved conflicts

arttemiuss43:
I can't make it work. I add `.env` file in the root of a project with content `GRIDSOME_API_URL=http://localhost`. Then when I use it in component like `process.env.GRIDSOME_API_URL` it returns undefined. I also tried to add environment variables in env config in gridsome.config.js and add environment based env files like `.env.development`. Nothing works. It only works when I use following in gridsome.config.js:

```
if (process.env.NODE_ENV === 'production') {
    require('dotenv').config({ path: '.env.production' })
} else {
    require('dotenv').config({ path: '.env' })
}

module.exports = {
    chainWebpack: config => {
        config
            .plugin('env')
            .use(require.resolve('webpack/lib/EnvironmentPlugin'), ['GRIDSOME_API_URL'])
    },
}
```

What am I doing wrong?

isoppp:
@arttemiuss43
This PR was merged, but didn't released yet.
Are you using master branch of this repository directly?
If you aren't using, this PR probably will include on v0.5.

And that feature don't need config.
You put `.env` or `.env.{mode}` file only then it's should work.

arttemiuss43:
Thank you for clarification. No, I don't. I just add gridsome as usual package using yarn.

isoppp:
@arttemiuss43 
Also, I tried your code, it's worked.

steps to reproduce:

1. exec `gridsome create xxx`
2. add `.env` file to root dir
```
GRIDSOME_API_URL=example
```
3. `yarn add dotenv`
4. add code writing above to gridsome.config.js
5. add `console.log(process.env.GRIDSOME_API_URL)` to `mounted()` function

noso2k1:
@arttemiuss43 did you maybe place by mistake the `.env` file in the `src`folder? This happened to me and always got `undefined`.
Make sure you're placing the file in the root, where for example `gridsome.config.js` is.

arttemiuss43:
Yeah. I placed .env file in the root of a project. This feature wasn't released yet when I faced this issue. It's released for now. So I will give it another shot. Thank you:)

PR #132, opened 2019-01-16
Fixes errors with JSON and Location content types

### LINKED ISSUE #129: Contentful extension throws type error on startup
## Description

I tried to connect my contenful repo to gridsome. i have 2 different spaces and it throws a different error on each of them.

### Steps to reproduce

This is the content model:
bildschirmfoto 2019-01-15 um 09 00 05

The config i use:
```
{
    use: "@gridsome/source-contentful",
    options: {
        space: "mykeyq", // required
        accessToken: "mykey35835102ff4ffb6f9fc257", // required
        host: "cdn.contentful.com",
        environment: "master",
        typeName: "Contentful"
    }
}
```
versions i use:
"@gridsome/source-contentful": "^0.2.0",
"gridsome": "^0.4.3"

### Expected result

Gridsome develop should startup correctly with contentful connected as data source.
Btw. if i connect DatoCMS it works fine.

### Actual result

following error appears:
bildschirmfoto 2019-01-15 um 09 00 14

on the other space i get:
bildschirmfoto 2019-01-15 um 08 57 54

hagata:
Adds a check for `sys` on fields that return an object before passing them to `this.createReferenceField`. Because the Location and JSON content types return objects without `sys` we can simply return the `value`

- Fixes #129
- adds examples and docs to README

### DIFF
--- a/packages/source-contentful/README.md
+++ b/packages/source-contentful/README.md
@@ -30,7 +30,7 @@ module.exports = {
 @gridsome/souce-contentful currently works with all Contentful Content Types except the Rich Text [Beta] type.
 
 ### Location
-Contentful Location data is returned as JSON with `lat:` and `lon`. You will need to query the field name and each field in the GraphQL query
+Contentful Location data is returned as JSON with `lat` and `lon`. You will need to query the field name and each field in the GraphQL query
 ```
 <page-query>
 query Location{
--- a/packages/source-contentful/README.md
+++ b/packages/source-contentful/README.md
@@ -25,3 +25,44 @@ module.exports = {
   ]
 }
 ```
+
+## Contentful Content Types
+@gridsome/souce-contentful currently works with all Contentful Content Types except the Rich Text [Beta] type.
+
+### Location
+Contentful Location data is returned as JSON with `lat:` and `lon`. You will need to query the field name and each field in the GraphQL query
+```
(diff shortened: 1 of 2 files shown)


packages/source-contentful/index.js
```diff
@@ -76,7 +76,7 @@ class ContentfulSource {
           fields[key] = value.map(item =>
             typeof item === 'object' ? this.createReferenceField(item) : item
```
hjvedvik:
```suggestion
            typeof item === 'object' && typeof value.sys !== 'undefined'
              ? this.createReferenceField(item)
              : item
```

packages/source-contentful/README.md
```diff
@@ -24,4 +24,45 @@ module.exports = {
     }
   ]
 }
+```
+
+## Contentful Content Types
+@gridsome/souce-contentful currently works with all Contentful Content Types except the Rich Text [Beta] type.
+
+### Location
+Contentful Location data is returned as JSON with `lat:` and `lon`. You will need to query the field name and each field in the GraphQL query
```
hjvedvik:
```suggestion
Contentful Location data is returned as JSON with `lat` and `lon`. You will need to query the field name and each field in the GraphQL query
```

tomtev:
Thank you 😍

PR #133, opened 2019-01-16
feat(transformer-yaml): add transformer

### LINKED ISSUE #131: YAML support
## Summary
I would like to use yaml with gridsome.
so I made gridsome-transformer-yaml plugin as a local plugin.

https://github.com/engineers-lt/portal/tree/master/plugins/gridsome-transformer-yaml

Are you planning to support yaml?

### Basic example

```js
const jsYaml = require('js-yaml')

class YamlTransformer {
  static mimeTypes () {
    return ['text/yaml']
  }

  parse(content, options) {
    const yaml = jsYaml.load(content)

    return {
      fields: {
        yaml
      }
    }
  }
}

module.exports = YamlTransformer
```

This is plugin code.

```yaml
- title: title1
  description: desc
- title: title2
  description: desc2
```

We will convert this data as follows

```graphql
query {
  report: report(path: "/data/reports") {
    yaml {
      title
      description
    }
  }
}
```

```
report {
  yaml [
    {
       title: 'title1',
       description: 'desc'
    },
    {
       title: 'title2',
       description: 'desc2'
    }
  ]
}
```

### Motivation

It is preferable to make it in a format like yaml when trying to make it with less budget.

I would to submit Pull Request.

mottox2:
This PR add `@gridsome/transformer-yaml` plugin to gridsome.
This should close #131.

### DIFF
--- a/packages/transformer-yaml/README.md
+++ b/packages/transformer-yaml/README.md
@@ -1,4 +1,4 @@
-# @gridsome/transformer-remark
+# @gridsome/transformer-yaml
 
 > YAML transformer for Gridsome
 
--- a/packages/transformer-yaml/README.md
+++ b/packages/transformer-yaml/README.md
@@ -1,0 +1,7 @@
+# @gridsome/transformer-remark
+
+> YAML transformer for Gridsome
+
+## Install
+- `yarn add @gridsome/transformer-yaml`
+- `npm install @gridsome/transformer-yaml`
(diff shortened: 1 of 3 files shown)


hjvedvik:
Thanks :) Should the parsed `yaml` data be used as fields to make them accessible as `node.myField` etc. instead?

mottox2:
Thanks. I fixed schema.
If yaml is object, we can access as `node.myField`, but if yaml is array, we can access as `node.data`.

PR #143, opened 2019-01-22
feat(pager-class-props): Support for additional css class overrides

boedlen:
Added additional props to Pager component to allow css class overrides on generate links and nav element

PS: This is my first contribution to OSS, so let me know if there's anything I should do differently next time :D


hjvedvik:
Thanks! You have done a good job :) I think the `activeLinkClass` and `exactActiveLinkClass` properties will be useful for many. But the `navClass` property might be unnecessary because you can use `class="my-class"` to set a custom class on `nav` since it is the root element.

boedlen:
Ahh, I didn't know attributes were passed down to the component root node.
I'll remove the `navClass` prop

PR #153, opened 2019-01-27
Allow IntersectionObserver polyfill

### LINKED ISSUE #144: Safari does not prefetch pages linked with <g-link>
## Description

`<g-link>` component provides a way for the site to prefetch linked pages which speeds up navigation considerably. Unfortunately it seems that Safari on MacOS and Safari on iOS don't prefetch pages and therefore there is a delay when navigating around a site for the first time.

### Steps to reproduce

Open the network tab in Chrome devtools.
Visit the [gridsome.org](https://gridsome.org/docs/) docs page in Chrome.
Note that pages types and content are prefetched, these are js files named things like `component--blog-post.d08379fa.js` and `24.5f8a2c20.js`. There should be quite a few of them.
Scroll down the menu and observe in the network tab that the content files are loaded in without clicking on a link.

Open the network tab in Safari devtools.
Visit the [gridsome.org](https://gridsome.org/docs/) docs page in Safari
Note that pages types are not prefetched, there are only two js files starting with `component--`. There are only two content js files. Scrolling the menu does not load in any more files. 

### Expected result

The Chrome and Firefox behaviour is the expected one, prefetching should work.

### Actual result

Safari does not prefetch content from `<g-link>` components.

tanc:
This commit makes a small change to the g-link component so that it no longer uses caniuse to check for IntersectionObserver support and instead checks directly for it. This allows the user to load a polyfill if they wish to support Safari (and potentially other browsers). Fixes #144


tanc:
Added the same for the `g-image` component.

hjvedvik:
I get a JS error when trying this PR. But I have refactored the internal Vue app a bit and simplified the `IntersectionObserver` check. I refactored to let `main.js` and plugins be imported before anything else. That will let you import a polyfill in your `main.js`. I have tested with the `intersection-observer` package and it works great. Just had to wrap it in `process.isClient` because it refers `window`. The changes are in the `master` branch now if you want to test it :)

tanc:
Closing as the work now on master branch fulfils the need

PR #162, opened 2019-01-30
Allow index.html template to be overridden

K3TH3R:
This small PR basically allows someone to add an `index.html` to their `/src` directory and override the default template. It's probably not going to happen very often, but it's the easiest way to allow people to add CDN/external scripts to the footer of their site.

### DIFF
--- a/gridsome/lib/app/loadConfig.js
+++ b/gridsome/lib/app/loadConfig.js
@@ -21,6 +21,7 @@ module.exports = (context, options = {}, pkg = {}) => {
   const args = options.args || {}
   const config = {}
   const plugins = []
+  const localIndex = resolve('src/index.html')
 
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks! Will be great to have the posibillity to override the default template. We'll include this in the next release :)

physcocode:
+1

tomtev:
This is great! Could you also create documentation for it on gridsome.org?

K3TH3R:
@tomtev Absolutely. Is there a page you would prefer it added to or should I create a new page in the sidebar? I'm thinking maybe "Overriding Index.html" as an item right below "body & html attributes"?

K3TH3R:
Documentation for this added to my existing documentation PR

Issue #167, opened 2019-02-01
Feature Request: Service Worker (PWA)


ipranjal:
## Summary
I have seen gridsome has "service worker" in their roadmap for 1.0 here is a list of flow I propose to add service worker, this is exact steps I used to create service worker for my project

### Steps
1) Add workbox-cli as one of the npm dependencies
2) In the gridsome "create" command add a basic workbox-config.js in the created project.
Here is a workbox-config.js I use in my gridsome project
```javascript
module.exports = {
  "globDirectory": "dist/",
  "globPatterns": [
    "**/*.{html,css,js,png,json}"
  ],
  "swDest": "dist/sw.js"
};
```
3) Exexute `npx workbox generateSW workbox-config.js` during `gridsome build` command

Hurray now we have a gridsome project with a 100 PWA score. Here is a really basic manifest.json file that is enough for my project.

```json
{
    "name": "Gridsome Project",
    "short_name": "Gridsome",
    "icons": [
      {
        "src": "logo-192.png",
        "sizes": "192x192"
      },
      {
        "src": "logo-512.png",
        "sizes": "512x512"
      }
    ],
    "lang": "en-US",
    "start_url": "/",
    "display": "standalone",
    "theme_color": "#29b473",
    "background_color": "#FFF"
}
```
I have added this manifest.json and logo-512, logo-192 files inside my static folder.

hjvedvik:
I have a plugin in a local branch that does almost what you are doing. I don't have much experience with Service Workers, but I can push the branch so you can see if it will work :)

hjvedvik:
Pushed a [`plugin-pwa`](https://github.com/gridsome/gridsome/tree/feat/plugin-pwa/packages/plugin-pwa) branch now that you can test. Would be great if you have the time to finish it :)

physcocode:
On to it

physcocode:
Just ran into a small problem otherwise plugin is ready to be in the core. I have hosted a test build on https://testpwag.firebaseapp.com everything works fine. And here is the lighthouse test result

<img width="659" alt="screenshot 2019-02-01 at 11 22 14 pm" src="https://user-images.githubusercontent.com/7591484/52140555-20850e80-2679-11e9-9005-adf46621948e.png">

physcocode:
@hjvedvik you may close this issue once the plugin has been brought to master

darkrubyist:
The Firebase link doesn't work on my end.

physcocode:
[...]

PR #168, opened 2019-02-01
fix workbox-build version

physcocode:
4.x is currently on alpha/beta channel and trying to npm install this plugin causes the following issue

` No matching version found for workbox-build@^4.0.0`


physcocode:
closes issue #167 , once this fix is applied you may release plugin-pwa I have tested it and this works

hjvedvik:
Thanks for testing the plugin @physcocode Sorry for the delay on this... One thing that I think we should look at before publishing is how to detect if the site is updated. And provide a Vue component or methods to invalidate the cache. I haven't been able to look into it yet, but I can probably do it in a couple of weeks.

PR #70, opened 2018-12-11
[WIP] Add siteDescription configuration

Devportobello:
Configuration tell us we can set the meta description but this is a missing feature

Documentation:
https://gridsome.org/docs/config
https://deploy-preview-28--gridsome-org-live.netlify.com/docs/config

With this you set the siteDescription property :
`gridsome.config.js`
```js
module.exports = {
  ...
  siteDescription: "My super site"
  ...
}
```

WIP:
I'm stuck when using `__dangerouslyDisableSanitizersByTagID` property of vue-meta to not satanize content

PR #169, opened 2019-02-01
Added key to viewport

physcocode:
I wanted to override default viewport information, but I could not.

### DIFF
--- a/gridsome/app/head.js
+++ b/gridsome/app/head.js
@@ -21,7 +21,7 @@ const head = {
   meta: [
     { charset: 'utf-8' },
     { name: 'generator', content: `Gridsome v${config.version}` },
-    { name: 'viewport', content: 'width=device-width, initial-scale=1, viewport-fit=cover' },
+    { key:'viewport' name: 'viewport', content: 'width=device-width, initial-scale=1, viewport-fit=cover' },
     
     // do not convert telephone numbers
     // into hypertext links because it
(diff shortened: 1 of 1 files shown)


physcocode:
Also siteDescription do not work for me, does it have something to do with 
description: config.siteDescription not present in this file ?

I encourage siteDescription to be mandatory to have a higher SEO score

physcocode:
oops seems like #70  is yet to be merged

tomtev:
What does key: 'viewport' do? Also, I think you missed a comma :)

physcocode:
@tomtev well actually I wanted to change the viewport meta to content: 'width=device-width, initial-scale=1, shrink-to-fit' but I could not override this because there was no 'key' set

tomtev:
Ah, nice. Maybe we should add a key to charset also? 

We should probably also make a 'head' option for gridsome.config.js so we could add for example:
```
head: {
  { key:'viewport', name: 'viewport', content: 'width=device-width' },
}
```

physcocode:
@tomtev @hjvedvik eagerly waiting for #70 and this to be merged in the core :)

tomtev:
Yes, its easy when you know how to do it :) Was just thinking its simpler for users to read config docs and see that there is are options for adding stuff to head. But we have a good documentation here https://gridsome.org/docs/head so I agree it's probably not needed.

PR #170, opened 2019-02-01
Added placeholder to ClientOnly

physcocode:
The placeholder is displayed until Client only component is mounted, useful when mounting component takes time

Inspired from no-ssr used by nuxt

### DIFF
--- a/gridsome/app/components/ClientOnly.js
+++ b/gridsome/app/components/ClientOnly.js
@@ -1,13 +1,34 @@
 export default {
+  name: 'NoSsr',
   functional: true,
-
-  render (h, { parent, children }) {
-    if (parent._isMounted) {
-      return children
-    } else {
-      parent.$once('hook:mounted', () => {
-        parent.$forceUpdate()
-      })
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks, a placeholder would be great!

I have only a few thoughts about the `placeholder` slot. Maybe it would be an idea to always render it on the server. And prepended the `vnode` created from the `placeholder` prop to the list instead. That will prevent the `placeholder` slot from being wrapped in a `div` element. But it might also make the new props unnecessary because you can just have the tag and placeholder text as a slot instead. 

I also have to test if rendering two different slots will have any impact on hydration in the browser.

Or what do you think? :)

physcocode:
Secondly you can style your placeholder by clientonly class

physcocode:
Also refer to the pull request I have sent in docs for "external scripts" I have explained use of ClientOnly component there

hjvedvik:
Sorry for not answering you sooner, but we'll keep the `ClientOnly` component as it is for now.

PR #179, opened 2019-02-05
 Allow user to create project in current directory

thien-hoang:
It would be awesome if we could create a new project in the current directory:

```
gridsome create .
```


hjvedvik:
Yes, that would be awesome! Thanks :)

PR #222, opened 2019-02-25
Add Prettier and ESLint to pre-commit

YogliB:


hjvedvik:
Thanks! Would it be possible to use `eslint --fix` instead of `prettier`? I haven't used pre-commit hooks for linting before, so I don't know what's possible or not :)

YogliB:
@hjvedvik Yes, it's even possible to add Prettier rules to eslint.

I generally like to separate my concern when I use Prettier for styling and TSLint/ESLint for code quality.

PR #227, opened 2019-02-27
Adds support for (existing) webp images

nkoehring:
This adds `webp` suffix to the webpack rule for images.


hjvedvik:
Thanks :)

PR #236, opened 2019-03-01
Only intercept left clicks on links

EdBailey:
I often navigate documentation sites by right clicking to open a link in a new tab, but I found that this was unexpectedly causing the browser to navigate. This makes it very difficult to browse.

This pull request simply ignores any clicks that don't use the left mouse button.

I hope this is in the right place. It was the only code i could see that looked like it would be causing the behaviour I was seeing.

### DIFF
--- a/gridsome/app/entry.client.js
+++ b/gridsome/app/entry.client.js
@@ -21,6 +21,10 @@ document.addEventListener('click', event => {
   if (
     event.defaultPrevented ||     // disables this behavior
     event.which !== 1 ||          // not a left click
+    event.metaKey ||
+    event.altKey ||
+    event.ctrlKey ||
+    event.shiftKey ||
     $el === null ||               // no link clicked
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks! Yes, should be the right place :) It's a global click event to make html links use the Vue router for internal URLs. I did a few tests now and it only seems to happen in Firefox. But when I hold the `cmd` key to open in a new tab on my Mac, it still navigates to the page in all browsers. Maybe we should include [this](https://github.com/vuejs/vue-router/blob/3cad567feb078f704162c4bce38441f2b8fc215b/src/components/link.js#L107) line from `vue-router` too?

EdBailey:
That makes sense, as I am using firefox on this computer. 
You're right, I'll add that line in as well!

hjvedvik:
Works much better now. Thanks!

PR #255, opened 2019-03-08
number of cpus defaults to 1

pabloveliz:
physical-cpu-count depends on lscpu on linux, which is not standard in all distros (ie. alpine) and on error returns 0, which breaks the build process with a message "Expected `concurrency` to be a number from 1 and up, got `0` (number)".
This fix returns 1 if the number of cpus is 0 (may as well return cpus.length, but that would defeat the original intention)


hjvedvik:
Thanks! We can probably allow to set a custom CPU count with an env variable or something, but this PR will work for now I think :)

Issue #134, opened 2019-01-18
Cross-Origin Request Blocked when using custom host = "0.0.0.0"


kaceo:
## Description
When a custom host is used in gridsome.config.js, the client cannot connect to the graphql layer 

     Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at
     http://0.0.0.0:8080/___graphql. (Reason: CORS request did not succeed).[Learn More]

### Steps to reproduce

  1. Clone gridsome/gridsome-starter-markdown-blog on test machine (IP = 192.168.1.1)

  2. yarn upgrade

  3. edit gridsome.config.js to include "host: 0.0.0.0" 

  4. yarn develop
      Compiled successfully in 7199ms                                        
      Site running at:          http://0.0.0.0:8080/
      Explore GraphQL data at:  http://0.0.0.0:8080/___explore

  5. From second workstation, use browser to access 192.168.1.1
 
### Expected result

Home page of the test blog

### Actual result

Empty screen.
Console log shows: 
OPTIONShttp://0.0.0.0:8080/___graphql
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://0.0.0.0:8080/___graphql. (Reason: CORS request did not succeed).[Learn More]

kaceo:
Also to add the following observation:

      gridsome develUsage: develop [options]
      
      start development server
       
       Options:
         -p, --port <port>  use specified port (default: 8080 )
         -h, --host <host>  use specified host (default: 0.0.0.0)
         -h, --help         output usage information

So the specification intends for port and host to default to these values, but the actual source line
in file https://github.com/gridsome/gridsome/blob/master/gridsome/lib/app/loadConfig.js
line 55 56 are:

       config.host = args.host || localConfig.host || 'localhost'
       config.port = parseInt(args.port || localConfig.port, 10) || 8080

shows that default host is localhost and not 0.0.0.0

This appears to be the default host is actually localhost.   Perhaps that is why
the CORS error wasn't noticed when testing against localhost, but in a cross
machine environment would be flagged.

kaceo:
Upon changing the gridsome.config.js file and set host = 192.168.1.1 the rest of code works.
So the original strategy of setting host to localhost only should also work, for environment where
server = client = localhost.

So the simple solution, to be added clearly into the documentation, is to put the actual IP into the gridsome.config.js and not to use "0.0.0.0".  This means a source repo may have a fixed IP in the config.

[...]

PR #271, opened 2019-03-11
allow OPTIONS method request to /___graphql

shooontan:
#134 

When using Docker, I set host option. Unfortunately Gridsome does not work. This problem is caused by that express-graphql accepts only `GET` or `POST` http methods. This PR resolves by accepting `OPTIONS` method in Gridsome's graphql middleware.

### before
command: `gridsome develop -h 0.0.0.0`
url: `localhost:8080`
result:
スクリーンショット 2019-03-11 18 47 29

### afrer
command: `gridsome develop -h 0.0.0.0`
url: `localhost:8080`
result:
スクリーンショット 2019-03-11 19 01 20

### DIFF
--- a/gridsome/lib/server/middlewares/graphql.js
+++ b/gridsome/lib/server/middlewares/graphql.js
@@ -9,6 +9,11 @@ const {
 
 module.exports = ({ store }) => {
   return async function (req, res, next) {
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thank you :)

PR #279, opened 2019-03-13
Support repeated path segments in dynamic routes

### LINKED ISSUE #277: `_raw` route params shouldn't be converted to string
## Description

[`path-to-regexp`](https://github.com/pillarjs/path-to-regexp#compile-reverse-path-to-regexp) allows you to generate paths with non-string parameters such as arrays and integers. For example, dynamic path segments require the parameter to be an array:

```js
const toPath = pathToRegexp.compile('/:pathSegments+')
console.log(toPath({ pathSegments: ['a', 'b', 'c'] }))
// Output: '/a/b/c'
```

The problem right now is that Gridsome [converts every parameter to a string](https://github.com/gridsome/gridsome/blob/dea7e8ecf13fc82fdb422ecea8069c6e166cd163/gridsome/lib/app/ContentTypeCollection.js#L304), even the `_raw` versions that are [supposed to contain the original value](https://gridsome.org/docs/routing#route-params). This means that the current behavior is analogous to the following:

```js
console.log(toPath({ pathSegments: String(['a', 'b', 'c']) }))
// Output: '/a%2Cb%2Cc'
```

### Steps to reproduce

Setup an example content type:

```js
// gridsome.server.js

module.exports = function (api) {
  api.loadSource(store => {
    const examples = store.addContentType({
      typeName: 'Example',
      route: '/example/:pathSegments_raw+'
    })

    examples.addNode({
      fields: {
        pathSegments: ['a', 'b', 'c']
      }
    })
  })
}
```

### Expected result

Run the following query in the playground:

```graphql
{
  allExample {
    edges {
      node {
        path
      }
    }
  }
}
```

The expected response would be:

```json
{
  "data": {
    "allExample": {
      "edges": [
        {
          "node": {
            "path": "/example/a/b/c"
          }
        }
      ]
    }
  }
}
```

### Actual result

Gridsome turns the `pathSegments` array into a string (`'a,b,c'`) which then gets encoded by path-to-regexp (`'a%2Cb%2Cc'`), and returns the following response:

```json
{
  "data": {
    "allExample": {
      "edges": [
        {
          "node": {
            "path": "/example/a%2Cb%2Cc"
          }
        }
      ]
    }
  }
}
```

### Environment

<pre><code>Libs:
- gridsome version: 0.5.5
</code></pre>

vberlier:
This fixes #277. Routes now support repeated path segments. Repeated keys with an array value are handled as expected:

```js
const examples = store.addContentType({
  typeName: 'Example',
  route: '/path/:segments+'
})

const node = examples.addNode({
  fields: {
    segments: ['a', 'b', 'c']
  }
})

console.log(node.path)
// Output: '/path/a/b/c'
```

The elements of the array are slugified unless you use the `_raw` suffix. `['foo bar', 'HELLO']` would result in `'/foo-bar/hello'` with `/:segments+` and `'/foo%20bar/HELLO'` with `/:segments_raw+`.

Note that node references also work:

```js
const node = examples.addNode({
  fields: {
    segments: [
      { typeName: 'Test', id: 1 },
      { typeName: 'Test', id: 2 }
    ]
  }
})

console.log(node.path)
// Output: '/path/1/2'
```

I tried to keep the internal changes to a minimum. Every object in the `routeKeys` array now has a `repeat` property and the original name of the key. Since the `name` property was used to hold the name of the field I renamed it to `fieldName` and used `name` to store the original name of the key.

This means that in the loop that populates the `params` object, `name` now iterates over every key name instead of every possibly duplicated field name. This removes the need for the [`&& !params[name]`](https://github.com/gridsome/gridsome/pull/279/commits/9200334c0ad0dc7d1eed9eb2ce105a204655cfa4#diff-5b4067434c16268d094326a114df9a04L303) and makes it possible to get each path segment simply by mapping over the values of the field array.

### DIFF
--- a/gridsome/lib/app/ContentTypeCollection.js
+++ b/gridsome/lib/app/ContentTypeCollection.js
@@ -287,22 +287,34 @@ class ContentTypeCollection extends EventEmitter {
     // Param values are slugified but the original
     // value will be available with '_raw' suffix.
     for (let i = 0; i < length; i++) {
-      const { name, path } = routeKeys[i]
-      const value = get(node, path, name)
+      const { name, path, fieldName, repeat } = routeKeys[i]
+      const field = get(node, path, fieldName)
 
-      if (name === 'year') params.year = date.format('YYYY')
-      else if (name === 'month') params.month = date.format('MM')
-      else if (name === 'day') params.day = date.format('DD')
-      else if (
-        isPlainObject(value) &&
-        value.hasOwnProperty('typeName') &&
-        value.hasOwnProperty('id') &&
-        !Array.isArray(value.id)
-      ) {
-        params[name] = String(value.id)
-      } else if (!isPlainObject(value) && !params[name]) {
-        params[name] = this.slugify(String(value))
-        params[name + '_raw'] = String(value)
+      if (fieldName === 'year') params.year = date.format('YYYY')
+      else if (fieldName === 'month') params.month = date.format('MM')
+      else if (fieldName === 'day') params.day = date.format('DD')
+      else {
+        const repeated = repeat && Array.isArray(field)
+        const values = repeated ? field : [field]
+
+        const segments = values.map(value => {
+          if (
+            isPlainObject(value) &&
+            value.hasOwnProperty('typeName') &&
+            value.hasOwnProperty('id') &&
+            !Array.isArray(value.id)
+          ) {
+            return String(value.id)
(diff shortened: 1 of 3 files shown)


hjvedvik:
This looks very good! I'll check it out tomorrow :)

hjvedvik:
Works great :) Thanks for the PR!

Issue #237, opened 2019-03-02
markdown excerpts 


roborich:
## Summary
Currently the @gridsome/transformer-remark plugin extracts an `excerpt` field from gray-matter parser which is always empty.  

### Basic example
https://github.com/gridsome/gridsome/blob/48bf6004de6ee578eeea40871a0dbedb034a923a/packages/transformer-remark/index.js#L49

The `exerpt` field is always empty because no excerpt options are passed to the gray-matter function. (Options are passed as a second parameter. See https://www.npmjs.com/package/gray-matter#options)

This should either be removed or the excerpt functionality should be added.  As the `except` field appears in the graphql schema, and can never be used.

### Motivation

I can currently stand to benefit from excerpt functionality.  Specifically, I could benefit from having a small portion of my .md file's content pre-formatted, as I will be showing exerpts of posts in a "recent posts" section on my site.

I haven't dug into the code enough yet to know how we would pass excerpt options in, but I am willing to help contribute to this if it's something that would be helpful.

Currently gray-matter has two excerpt options
```js
exerpt // Boolean|Function
excerpt_separator // String
```
we could offer those as direct options to the remark plugin, or simplify gray-matter's somewhat wonky excerpt function api and allow a simple transform function to be used.  Where someone could pass something like...

```js
{
  transformers: {
    remark: {
      createExcerpt: (rawContent) => rawContent.split('\n')[0];
      // ... 
    },
  },
}
```

roborich:
Made a PR #301

apvarun:
Any update on this PR?

jlabs:
Also interested in an update? 
Was hoping to use the excerpt field on a blog I'm making - have an idea of a workaround, but it's too hacky for my likings.

sangeet:
You can use this function on any content field and generate excerpt during runtime. Strips out the HTML tags and extracts only the fiirst n characters.

Not the cleanest way to do things, but it's a handy utility.

```
    function generateExcerpt(textToStrip, length) {
      let regex = /(<([^>]+)>)/gi
      return textToStrip.replace(regex, "").slice(0,length)
    },
```

noxify:
Here an alternative way to create the excerpt:

```js
//gridsome.server.js

  api.loadSource(({addSchemaResolvers}) => {
    addSchemaResolvers({
      Article: {
        excerpt(obj) {
          var longText = (obj.excerpt.length > 200) ? '...' : '';
[...]

PR #284, opened 2019-03-16
transform-remark: front-matter parser config

eostrom:
Allow configuration of `gray-matter` from `gridsome.config.js`.


eostrom:
Specifically I'm trying to use the `excerpt` and `excerpt_separator` options to parse Markdown posts from my old blog, but it seemed like the best approach was just to take an object of options and pass them through to `gray-matter`.

roborich:
By the looks of it, this will solve the issue I created regarding remark.  #237  

I really think that the transform-remark plugin needs to either have `excerpt` functionality **OR** it needs to remove `excerpt` from the schema.  It is currently an unusable field in the GraphQL schema and is fairly confusing.

robaxelsen:
Closing, as [PR with alternative solution](Prefixing the slugs which have leading numbers is safer, but results in uglier slugs.) has already been merged.

eostrom:
@robaxelsen The PR you linked to was also closed without merging. Not saying that this one has to stay open, but has another solution for Markdown excerpts been implemented? It's not mentioned in the documentation.

robaxelsen:
@eostrom Oh dang, my bad! Really sorry about that, and the confusion caused. Reopening!

robaxelsen:
Oh wait.. Seems I assumed this commit had implemented this: https://github.com/gridsome/gridsome/pull/284/commits/9245c929b40c2693c9873e4a677ba1c1d89c179f

@hjvedvik is that the case?

@eostrom please test if you have the chance

eostrom:
That commit brought my branch up to date with master (as of a week ago), but didn't merge my branch *into* master.

I can confirm that, with those updates, my branch still passes options to `gray-matter` the way I expect it to.

However, the excerpt is not parsed as Markdown. I can't remember whether this worked previously. I suspect not. (I've been going without excerpts for now.) Without Markdown parsing, this feature isn't useful to me personally. My use case is essentially #237.

At this point, I think the options are:

- Close this.
- Merge it, as the ability to configure `gray-matter` might be useful to someone.
- It looks to me like @hjvedvik has a plan for solving #237, which may make this moot.
- Given guidance on how the team thinks #237 should be solved, I'd be happy to help implement it.

I am content with any of these.

hjvedvik:
@eostrom I think `gray-matter` only extracts the excerpts but we probably have to add an `excerpt` GraphQL field which parses the excerpt with the Remark parser. It would work pretty much like the `content` GraphQL field.

I wonder if it would be better to have an `excerpt` option instead of `grayMatter` and allow it to be a boolean or a function. Or what do you think?

hjvedvik:
Thanks for the PR @eostrom :) And sorry for not merging it sooner... The plan has been to deprecate transformers and have a source plugin for each file type instead. But we haven't got the chance to start on them yet.

PR #293, opened 2019-03-18
feat(transformer-csv): add csv transformer

### LINKED ISSUE #282: Feature request: Transformer for CSV
## Summary

A transformer similar to json and yaml, but for csv.

### Basic example

```javascript
const csvParse = require('csv-parse/lib/sync')

class CSVTransformer {
  static mimeTypes () {
    return ['text/csv']
  }

  parse (source) {
    const records = csvParse(source, {
      columns: true,
      skip_empty_lines: true
    })

    const fields = { records }

    return { fields }
  }
}

module.exports = CSVTransformer
```

This would parse a CSV source using csv-parser, using the first line for columns names. The parser creates an array of record objects. Each object has keys-value pairs corresponding to each entry's column name and value. This array is then returned as part of fields.

The records can then be queried like so
```
allFileNode {
  edges {
    node {
      records {
        column 1
        column 2
        column 3
      }
    }
  }
}
```
### Motivation

This will allow CSV data sources to be added and queried.

arjunchib:
Adds `@gridsome/transformer-csv` plugin to parse `csv` sources. Closes #282.

### DIFF
--- a/packages/transformer-csv/README.md
+++ b/packages/transformer-csv/README.md
@@ -1,0 +1,8 @@
+# @gridsome/transformer-csv
+
+> CSV transformer for Gridsome
+
+## Install
+
+- `yarn add @gridsome/transformer-csv`
+- `npm install @gridsome/transformer-csv`
--- a/packages/transformer-csv/__tests__/index.spec.js
+++ b/packages/transformer-csv/__tests__/index.spec.js
@@ -1,0 +1,61 @@
+const CSVTransformer = require('../index')
+
+test('parse valid csv', async () => {
+  const transformer = new CSVTransformer()
+  const res = transformer.parse(`header 1,header 2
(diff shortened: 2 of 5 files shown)


hjvedvik:
Thanks! This looks good :) Would it make more sense to have each line as its own node? It isn't possible atm, but we could try to make it work in core.

arjunchib:
Yeah thats probably the better way to do it. Might be good to have the option to do both? Could see a situation where someone has a collection of CSVs where each one is its own page.

hjvedvik:
We can try to allow the `parse` method to return an array of node options in the next version of Gridsome (0.6). I think it could be useful for other transformers as well. This plugin can probably be like it is now and you can add an option for multiple nodes per file once the required changes are in core.

PR #295, opened 2019-03-19
Fix field name and suffix separation

### LINKED ISSUE #294: Route with deep field starting with `raw` doesn't work (e.g.: foo__rawValue)
## Description

Referencing a deeply nested field in a route doesn't work if the field name starts with `raw`. This is because the field name is currently obtained from [`key.name.replace('_raw', '')`](https://github.com/vberlier/gridsome/blob/2259f85de11f0b0139d2c0cfac87c6be0217ae35/gridsome/lib/app/PluginStore.js#L97). Separating the field name from the suffix with a regular expression would fix the problem.

### Steps to reproduce

```js
const contentType = store.addContentType({
  typeName: 'TestPost',
  route: '/:foo__rawValue'
})

const node = contentType.addNode({
  fields: {
    foo: {
      rawValue: 'BAR'
    }
  }
})
```

### Expected result

```js
console.log(node.path) // Output: '/bar'
```

### Actual result

```js
console.log(node.path) // Output: '/foo-value'
```

### Environment

```
gridsome version: 0.5.5
```

vberlier:
This fixes #294. Deeply nested field references will now work properly if the field name starts with `raw`. The field name and the suffix are now separated with a regular expression instead of removing every occurrence of `_raw` in the key with [`key.name.replace('_raw', '')`](https://github.com/vberlier/gridsome/blob/2259f85de11f0b0139d2c0cfac87c6be0217ae35/gridsome/lib/app/PluginStore.js#L97).

```js
const contentType = store.addContentType({
  typeName: 'TestPost',
  route: '/:foo__rawValue'
})

const node = contentType.addNode({
  fields: {
    foo: {
      rawValue: 'BAR'
    }
  }
})

console.log(node.path) // Output: '/bar'
```

Every element of the `routeKeys` array now has an additional `suffix` property that holds either the value of the suffix or `null` if the key doesn't have any suffix.

### DIFF
--- a/gridsome/lib/app/ContentTypeCollection.js
+++ b/gridsome/lib/app/ContentTypeCollection.js
@@ -287,7 +287,7 @@ class ContentTypeCollection extends EventEmitter {
     // Param values are slugified but the original
     // value will be available with '_raw' suffix.
     for (let i = 0; i < length; i++) {
-      const { name, path, fieldName, repeat } = routeKeys[i]
+      const { name, path, fieldName, repeat, suffix } = routeKeys[i]
       const field = get(node, path, fieldName)
 
       if (fieldName === 'year') params.year = date.format('YYYY')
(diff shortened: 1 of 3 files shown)


hjvedvik:
Thanks again! Looks like a good solution :)

PR #297, opened 2019-03-19
Support root option

morrislaptop:
unknown

Example usage:

```
    {
      use: "@gridsome/source-filesystem",
      options: {
        path: "pages/**/*.yml",
        typeName: "ComponentPage",
        root: 'pages',
      }
    },
```

### DIFF
--- a/packages/source-filesystem/index.js
+++ b/packages/source-filesystem/index.js
@@ -9,6 +9,7 @@ class FilesystemSource {
     return {
       path: undefined,
       route: undefined,
+      root: undefined,
       index: ['index'],
       typeName: 'FileNode',
       refs: {}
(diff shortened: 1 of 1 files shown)


hjvedvik:
We are going to improve the options for the filesystem source plugin. And beeing able to define a base folder is one of the things we wanted to look at. But I'm not sure if replacing the dir path is the best solution. I think we should try to change the value of `this.context` instead. You can join the value of `api.context` and `options.root` and use it as `this.context`.

And maybe the options should be named `baseDir` instead?

morrislaptop:
No opinion on the means, just need the end /about-us/team :)

hjvedvik:
We've added a `baseDir` option to the filesystem plugin which solves this issue. It will probably be published tomorrow so I'm closing this :)

PR #301, opened 2019-03-20
transformer-remark excerpt functionality

roborich:
Adds a `createExcerpt` option for transformer-remark plugin.

`excerpt` is currently in the GraphQL schema for any data created by the transformer-remark plugin.  But it is **always empty** because the plugin never passes excerpt options to gray matter.

This addresses an issue (#237) I created, and is an alternate solution to PR #284 with a more simplified API that doesn't require a user to know the gray matter package's options and their less simple excerpt function.

`createExcerpt` is just a function that accepts the raw markdown as an argument, and returns the excerpt.

### DIFF
--- a/packages/transformer-remark/README.md
+++ b/packages/transformer-remark/README.md
@@ -15,6 +15,7 @@ module.exports = {
       externalLinksTarget: '_blank',
       externalLinksRel: ['nofollow', 'noopener', 'noreferrer'],
       anchorClassName: 'icon icon-link',
+      createExcerpt: rawContent => rawContent.split('\n')[0],
       plugins: [
         // ...global plugins
       ]
@@ -45,6 +46,11 @@ module.exports = {
 - Type: `array` Default: `[]`
 
 Add additional plugins to the parser.
+
+#### createExcerpt
+- Type: `function` Default: `undefined`
+
+Creates an excerpt from the raw markdown content.
 
 #### useBuiltIns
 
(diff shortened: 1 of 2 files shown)


roborich:
Also, I see that there are default options passed into the plugin, but I couldn't find where they are coming from.  The code could be cleaned up a bit if there was a no-op (`() => undefined`) as the default value for the `createExcerpt` function.

hjvedvik:
Thanks! The `content` and `excerpt` fields will be removed from the `NodeInterface` in 0.6, but it would be great if the Remark transformer added an `excerpt` field :) Should the option be named `excerpt` and allow it to be `true` OR a function instead?

roborich:
:+1: to it being removed from the `NodeInterface`.  It's much better than always empty.  

If the `excerpt` option was a boolean, what would the excerpt include?  A pre-set number of words or lines? Maybe the option could be a number, for the number of words.  If it was a number we could allow it to be a field argument in GraphQL instead of a config option, which honestly sounds a lot better because it will allow for much more flexibility.

I think it would be really helpful to add the [`remove-markdown`](https://www.npmjs.com/package/remove-markdown) package (or similar) to the plugin to strip out markdown syntax from the excerpt.

hjvedvik:
`excerpt: true` should just use the default behavior mentioned in the [docs](https://www.npmjs.com/package/gray-matter#optionsexcerpt):

> If set to `excerpt: true`, it will look for the frontmatter delimiter, `---` by default and grab everything leading up to it.

We are going to let transformers parse field values too, but I'm not sure when we'll implement it yet. So we could just support plain text as excerpt for now. Or transform it to HTML in the `parse` method to allow bold, and italic etc. using the `toAST` and `toHTML` methods.

roborich:
I'm going to close this.  When the time comes to update the plugin, it will be done differently than what is in this branch.

PR #308, opened 2019-03-22
Add documentation on how to add custom routes

ameistad:


hjvedvik:
Thanks :) Should the default options be excluded in that example? Just to make the `routes` option more visible.

robaxelsen:
@ameistad have you had a chance to consider the suggestion by @hjvedvik above? :point_up:

ameistad:
Sorry, was away for a couple of days. Removed the default options in the new pull request.

Issue #142, opened 2019-01-22
Source Request: Ghost


gottlike:
## Summary

Ghost (https://github.com/TryGhost/Ghost) recently announced it can be used as a fully headless CMS and they even have a Gatsby starter + plugin: https://blog.ghost.org/jamstack/

Therefore it would be great to have a Ghost source plugin for Gridsome!

### Basic example

Ghost content API: https://docs.ghost.org/api/content/

Gatsby source: https://github.com/tryghost/gatsby-source-ghost

### Motivation

I very much like Ghost as a CMS, but I don't particularly like React - so Gridsome + Ghost would be great!

technoSpino:
I think this can be closed

gottlike:
You're right - thanks :)

PR #318, opened 2019-03-25
Add Ghost Datasource

technoSpino:
Resolves issue #142


hjvedvik:
Thanks for the PR :)

I see it uses the `normalizeFields` and `normalizeFieldValue` methods from the WordPress source plugin. But I don't think it's necessary to use them in this plugin because the posts don't have a `meta` or `post_type` property.

technoSpino:
Sorry about the overzealous C&P. Thought it was important for graphql. Updated and removed some of the unnecessary logging

hjvedvik:
No problem :) The methods were for creating references to other types. How will that work for this plugin? How is a post connected to its author for example? The Ghost documentation says related data like author etc isn't included unless an `include` option is specified.

The `adminKey` option isn't used, so I think you can delete it from the default options. And maybe `contentKey` can be renamed to `key`?

The default `routes` option is a mixture of singular and plural type names. Can you change them to only have singular type names?

And the `Usage` section in the readme file should use options related to this plugin. `baseUrl` should be `url` and it should also show the `key` option because it's also a required option.

hjvedvik:
Thanks for the updates :) I did a few changes + added support for settings as metadata.

Gridsome currently camel cases custom field names, but we are going to disable it by default in v0.6. Do you think this plugin should keep camel-cased names or maybe it's better to just stick with the same names as in the RESTful API?

technoSpino:
I would say have a pattern for gridsome and then have the plugins translate to each individual api

robaxelsen:
Hi wonderful people :smile: Not really clear to me from the comments who we're waiting for here. @hjvedvik @technoSpino: please reply with an update.

hjvedvik:
@robaxelsen @technoSpino I think it's ready to be merged. But I can update it to work with 0.6 first :)

hjvedvik:
@technoSpino I did some updates, but haven't been able to test. My trial period ended. Are you able to test the plugin with the latest changes?

dominiquedutra:
So I did some testing. Hope it helps.

Self hosted ghost (@hjvedvik - you don't need a paid trial to test it, ghost can be self-hosted and it's as easy as installing it and doing a ghost start and going to localhost:2368 > integrations and setting a custom api key).
I was able to pull and list the posts just fine. All data seem to be populated correctly on graphql.

Some minor things @technoSpino 

1) options on packages/source-ghost/README.md are documented wrong. README says `baseUrl` should be set while index.js expects `url`. Also, you must set a `contentKey` as `key` and this is not documented on README.md for the source. I suggest using baseUrl and contentKey as options considering there is also an adminKey.

2) Build failed if my posts do not have tags. Had to comment out lines 106 to 109 on index.js.

3) I guess the graphql queries should be renamed to plural (allGhostPost > allGhostPosts)?

Other than that it works. Thank you very much.

PS: I am just starting out on gridsome. I could submit some code but I am not yet confident.

hjvedvik:
Thanks for your feedback @dominiquedutra. Just installed Ghost locally and it works great! I'll change `url` and `key` back to `baseUrl` and `contentKey` and check if the post has a `primary_tag` :)

technoSpino:
Thanks for the help @hjvedvik and @dominiquedutra

PR #319, opened 2019-03-26
Modify Article query so image url resolves

jfitzsimmons2:
This would fix a small piece of the [Drupal plugin docs](https://github.com/gridsome/gridsome/tree/master/packages/source-drupal#example-page-queries). Currently, the single Article query example throws an error, but if you change the GraphQL query to what's in the PR, it returns the file URL in fieldImage.

Before:
Screen Shot 2019-03-25 at 10 44 57 PM

After:
Screen Shot 2019-03-25 at 10 45 20 PM


hjvedvik:
Thanks :)

PR #344, opened 2019-04-02
Add link range option to pager

sanscheese:
Have an option to set a `linkRange` (number of pages/links to display).

Also, set the default to 5 as 10 is quite high.


hjvedvik:
Nice :) Do you think it would be easier to just call it `range`?

sanscheese:
Was having similar thoughts on naming. Was considering `pageRage` following other vue pagination library. Have changed to `range`.

hjvedvik:
Thanks! I think it looks good now. Will include this in 0.6 :)

PR #345, opened 2019-04-03
Update index.js for featured images and post modified

joeelia:
Posts with featured images now actually show featured images when using gridsome build command. The data structure is now 

featuredMedia {
      url
      width
      height
      altText
    }

Also added date modified for posts incase someone has a post that they keep modifying over time, this is good for search engines to know you post is still relevant.

### DIFF
--- a/packages/source-wordpress/index.js
+++ b/packages/source-wordpress/index.js
@@ -109,14 +109,11 @@ class WordPressSource {
       }
     }
   }
-  
 async getFeaturedMedia (featured_media) {
-  const image = await this.fetch('wp/v2/media/', {featured_media}, {})
-  const fields = this.normalizeFields(image.data[0])
-  return fields;
-    
+    const image = await this.fetch('wp/v2/media/', { featured_media }, {})
+    const fields = this.normalizeFields(image.data[0])
+    return fields; 
 }
-  
   async getPosts (store) {
     for (const type in this.restBases.posts) {
       const restBase = this.restBases.posts[type]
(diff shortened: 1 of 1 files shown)


hjvedvik:
This change will run one API request per post and will increase the time to run `loadSource` a lot. It would be better to fetch all media posts and then create a reference with the `featuredMedia` field. But the current implementation already does that in v0.2.1. The following query works:

```gql
query Posts {
  allWordPressPost {
    edges {
      node {
        modified # post last modified
        featuredMedia {
          sourceUrl # url to featured image
          mediaDetails {
            width
            height
          }
        }
      }
    }
  }
}
```

joeelia:
Ah okay I wasn't aware you have already accounted for this because with this current index.js causes this template to have errors / not work correctly: https://github.com/gridsome/gridsome-starter-wordpress/blob/master/src/templates/WordPressPost.vue, but thats a separate issue. This script is fine now that I know the proper GraphQL query.

This can be closed.

hjvedvik:
No problem :) There is a PR open in the starter repo that fixes the query and updates the source plugin to latest version. It should be ready to merge pretty soon.

jacobarriola:
@hjvedvik out of curiosity, I can't seem to figure out how the `featuredMedia` object is getting its data (`sourceUrl`, `mediaDetails`). The WordPress API only makes the media id available, so I've been looking through the source code to figure out where a subsequent request is made.

The only thing that comes close is https://github.com/gridsome/gridsome/blob/master/packages/source-wordpress/index.js#L140, but that doesn't seem to make an explicit call.

Again, I'm just being curious as to how some of the inner workings are put together - nothing is actually broken. 

Thanks again for making this. Awesome work. 👏🏽👍

hjvedvik:
@jacobarriola Yes, `post.featured_media` is just an id. The `createReference` method helps Gridsome to locate the attachment with that id. Which makes the GraphQL schema able to return the correct image data instead of just an id :)

PR #346, opened 2019-04-03
Refactor node options and NodeInterface

hjvedvik:
What this PR does:

- Removes `title`, `date`, `path`,  `slug`, `content` and `excerpt` from `NodeInterface`.
- Renames `NodeInterface` to `Node`.
- Accepts custom node fields as root options.

#### Custom fields

Custom fields should be added as top-level options to be more consistent with how the structure is in the GraphQL schema. Having custom fields under `fields` will not break anything, but it will be deprecated until v1.0 so you should update your code.

```diff
collection.addNode({
  title: 'A node',
- fields: {
-   customField: 'value'
- }
+ customField: 'value'
})
```

#### NodeInterface

The `slug`, `content` and `excerpt` fields are not used for all sources and it's unnecessary to have empty fields in the schema. Some source plugins also need to have something else than a string for the `content` field, and that's not possible as long as the interface wants it to be a string. This is a breaking change for queries which use them on the `NodeInterface` type. And `NodeInterface` is also renamed to `Node`.

We could also remove `title` and `date`, but that can be up to discussion. Currently, `:year`, `:month` and `:day` relies on the `node.date` property and it's used for default sorting. And the `path` field could be a reserved custom field for nodes which have a path or a route instead of being defined on the interface. Because not all node types are meant to have a path.

```diff
query {
  post {
    belongsTo {
      edges {
        node {
-         ... on NodeInterface {
+         ... on Node {
            id
-           title
-           path
-           slug
-           content
-           excerpt
          }
+         ... on Post {
+           title
+           path
+           slug
+           content
+           excerpt
+         }
        }
      }
    }
  }
}
```


onecrayon:
From an architectural standpoint, I think the main question is: Is Node intended as the common interface for absolutely all data types stored by Gridsome, or does it explicitly represent data for a specific route on the site?

If the former, then I would argue that decoupling `date` and `path` from the Node interface is highly desirable. However, I'm not sure exactly what the use-case would be for this off the top of my head, because it destroys (or at minimum complicates) most of the built-in sorting logic for Node collections. I guess on the plus side, it allows much more free-form entry of data into the system, which could make handling edge cases easier with custom plug-ins.

If the latter, then I would argue that including `date`, `path`, and possibly `slug` makes total sense on the root Node interface, although it might be worth discussing with the community whether folks would find it useful to be able to easily populate the GraphQL layer with custom types that do not implement the Node interface (maybe there's a use case that's somewhere between data that lives at a specific path, and static data that lives in `metaData`; I'm not sure).

Personally, I'm a big fan of moving `fields` custom fields to the root type, at least. :)

Lastly: given that you've moved `slug`, `content`, and `excerpt` out of the Node interface, what is the reason for maintaining `title`? I'm not against it, per se, but I can more easily imagine scenarios where I would want `title` to be optional than where `path` or `date` is optional (e.g. a microblog, image blog, etc. where the only pertinent data for a route/node is the timestamp and the content).

hjvedvik:
@onecrayon Thanks for your feedback :) I thought of nodes kind of like posts in WordPress or nodes in Drupal, where each node has a path where it will be available (if it has a template). So I think `title`, `path` and `date` are the most basic fields for this use case.

We could implement another data type for use cases where you don't need to render pages or want full control yourself with `api.createPages`. Kind of like entities in Drupal. That could also allow us to add more features to nodes (like `belongsTo` or previous / next node etc.) while entities are kept as simple as possible. And nodes could also inherit the entity interface.

onecrayon:
That all makes sense to me. Given that I'm having trouble thinking of a specific use-case off the top of my head for a lower level "entity", it might well be worth it to leave abstracting that level of functionality off for now, too (unless you've got feedback from other folks who need that sort of low-level data insertion without the overhead of collections and so forth).

If you've got a method for inserting pages associated with a path, and a method for inserting nodes associated with paths that automatically receive collections and previous/next support, that pretty much covers my normal use-cases.

hjvedvik:
@onecrayon I removed `title`, `date` and `path` from the interface too. They are currently handled in a special way some places in the code, but I think we'll move that to a separate plugin sometime in the future.

PR #350, opened 2019-04-05
Global 'click' event listener should account for port

Strato:
My gridsome website is deployed on `example.org`. I have a link pointing to `//examples.org:8080/foobar`.

This global 'click' event should account for the port in its test, otherwise, it mistakenly thinks this is an internal link and overrides the domain.

N.B: honestly this whole event handler seems like a terrible idea... It's easy to circumvent by right clicking on an anchor and do "Open in a new..." (tab, window, whatever) and I can see a lot of edge cases arise. People should just use `g-link`. But that's just my two cents.


hjvedvik:
Thanks! The event is for capturing internal links inserted by for example `v-html`. It just does the same checks as `router-link` plus a few more. But we might remove it if we find a better solution sometime.

PR #354, opened 2019-04-08
Adds Rich text content type support

hagata:
- Adds a handler for data coming from Contentful
- Adds documentation to the readme on how to render the Rich text to Vue

Reference for Contentful Rich text response and handling, 
https://www.contentful.com/developers/docs/tutorials/general/getting-started-with-rich-text-field-type/


packages/source-contentful/index.js
```diff
@@ -80,6 +80,9 @@ class ContentfulSource {
           )
         } else if (typeof value === 'object' && typeof value.sys !== 'undefined') {
           fields[key] = this.createReferenceField(value)
+        } else if (typeof value === 'object' && value.nodeType === 'document') {
+          // value is Rich Text
+          fields[key] = JSON.stringify(value)
```
hjvedvik:
```suggestion
          fields[key] = documentToHtmlString(value)
```
Would it be possible to generate the HTML here instead?

hagata: It's possible, but because the `documentToHtmlString` method takes a config option, we would have to pass that in through `gridsome.config.js` and then it becomes less customizable if, for instance, you needed to render custom elements for different pages. Like how in the readme we pass a custom config for handling embedded image assets

hjvedvik: I understand :) Maybe the rich text fields could contain two properties: `document` and `html`? The `html` property could be used for people who just want to show the HTML.

hagata:
@hjvedvik I've updated this with your suggestion to add html 🙌

hjvedvik:
Great :) I don't mean to be picky here, but what about using `contentType.addSchemaField()` to create a `GraphQLJSON` field? Then there will be no need to use `JSON.parse()` in the browser.

sebastianwestberg:
I think this feature is awesome and is quite essential for working with Contentful. Please shout if you need help in any way!

hagata:
Thanks for the update @hjvedvik I think if we just update the readme, it'll be ready to go.

hjvedvik:
@hagata I tried to update the readme, but feel free to make it better if you want to.

robaxelsen:
Status here? Waiting for @hagata to respond, otherwise ok from your point of view @hjvedvik?

PR #369, opened 2019-04-17
Add apiKey to response fields in DatoCMS

gilesbutler:
In graphQL it'd be great to access the `api_key` prop that Dato's JS lib provides.

The original field can be found in the [documentation](https://www.datocms.com/docs/content-management-api/resources/field).

It's very useful when using Dato's [Modular Content Blocks](https://www.datocms.com/docs/content-modelling/modular-content) as it helps identify which block type to use on the frontend.

Thanks

### DIFF
--- a/packages/source-datocms/index.js
+++ b/packages/source-datocms/index.js
@@ -109,7 +109,7 @@ class DatoCmsSource {
           const val = item.readAttribute(field)
 
           if (item.itemType.hasOwnProperty('apiKey')) {
-            fields.apiKey = item.itemType.apiKey
+            fields.model = { apiKey: item.itemType.apiKey }
           }
 
           if (!val) return fields
--- a/packages/source-datocms/index.js
+++ b/packages/source-datocms/index.js
@@ -107,6 +107,11 @@ class DatoCmsSource {
         updated: new Date(item.updatedAt),
         fields: item.itemType.fields.reduce((fields, field) => {
           const val = item.readAttribute(field)
+          
+          if (item.itemType.hasOwnProperty("apiKey")) {
+            fields.apiKey = item.itemType.apiKey;
+          }
+          
           if (!val) return fields
 
           if (['link', 'links', 'rich_text'].includes(field.fieldType)) {
--- a/packages/source-datocms/index.js
+++ b/packages/source-datocms/index.js
@@ -107,11 +107,11 @@ class DatoCmsSource {
         updated: new Date(item.updatedAt),
         fields: item.itemType.fields.reduce((fields, field) => {
           const val = item.readAttribute(field)
-          
-          if (item.itemType.hasOwnProperty("apiKey")) {
-            fields.apiKey = item.itemType.apiKey;
+
+          if (item.itemType.hasOwnProperty('apiKey')) {
+            fields.apiKey = item.itemType.apiKey
           }
-          
+
           if (!val) return fields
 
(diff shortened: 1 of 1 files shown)


hjvedvik:
@stefanoverna Do you have any comments on this one?

tomtev:
@stefanoverna poke :)

hjvedvik:
@gilesbutler Can the `__typename` field in the GraphQL schema also be used to identify block type?

And I noticed that `itemType.name` has been used to generate the typenames. Do you think `itemType.apiKey` would be more suited for the generated typename instead?

gilesbutler:
Hey @hjvedvik no I think it's correct to use `itemType.name` to generate typenames. You use the typename to get the fragment but then that fragment doesn't expose an identifier via the Gridsome Dato plugin. In the Dato JS SDK they return the `apiKey` prop so you can identify each sub section.

We use the typename to get the fragments data e.g `__typename` = `DatoProjectsSection` then...
```
      ... on DatoProjectsSection {
        title
        projects {
          id
          title
          description
          path
          featured
          featuredImage {
            url
          }
        }
      }
```

hjvedvik:
I meant that you can get the type with a `__typename` field in union fields like this:

```gql
query {
  sections {
    __typename # sets a __typename property on each item in the list
    ... on SomeType {
        title
    }
    ... on OtherType {
        title
    }
  }
}
```

I'm not sure if it's a good idea to generate type names from the human-readable names. The Gatsby plugin for DatoCMS uses `itemType.apiKey` as typeName instead of `itemType.name`. But we can probably fix that in another PR  :)

hjvedvik:
I've been looking a bit more on their Gatsby plugin. They have a `model` field on each node which has a reference to the item type. I don't know if the model types are fetched by this plugin, but we could set the `apiKey` field inside a `model` object for now. So you will have it at `node.model.apiKey`.

gilesbutler:
I see what you're saying with the `__typename` but we need the `__typename` at the next level down on the actual type itself e.g
```
... on SomeType {
        __typename
        title
    }
```

Having `node.model.apiKey` should be fine, as long as we have access to the `apiKey` at that level I think it will work thanks 👍

packages/source-datocms/index.js
```diff
@@ -107,6 +107,11 @@ class DatoCmsSource {
         updated: new Date(item.updatedAt),
         fields: item.itemType.fields.reduce((fields, field) => {
           const val = item.readAttribute(field)
+
+          if (item.itemType.hasOwnProperty('apiKey')) {
+            fields.apiKey = item.itemType.apiKey
```
hjvedvik:
```suggestion
            fields.model = { apiKey: item.itemType.apiKey }
```

gilesbutler:
Thanks @hjvedvik, ha I just tried `__typename` and it's working on each type now! I could have sworn it wasn't there before 🙈 anyway the `model.apiKey` will be super helpful thanks 👍

matjack1:
Hello, I'm chiming in to give a feedback from DatoCMS-land :)

I've checke the code and it makes sense to me to use the `api_key` as you are doing. I also think you can get rid of the if statement as the `api_key` must be present for every model to exist, should be safe, but it's up to you.

Thanks for doing this! If you need any further feedback let me know, I'm now following this PR.

hjvedvik:
Thanks for the feedback! @matjack1 :) I'll merge this. We can remove the if statement sometime later.

PR #389, opened 2019-04-28
Update sharp for supporting Node 12.x

### LINKED ISSUE #388: Add support for Node 12
Gridsome currently depends on `sharp` v0.21. But only `sharp` v0.22.1+ is compatible with Node 12, as it depended on some V8 internals that were removed (`v8::Handle`). Trying to install Gridsome on Node 12 fails with a `gyp` error.

For related issue upstream, see https://github.com/lovell/sharp/issues/1668.

leopiccionia:
This PR fix #388.


hjvedvik:
Thanks :)

PR #392, opened 2019-04-28
feat(sitemap): add support for static urls (gridsome#390)

ameistad:


hjvedvik:
Thanks :) Should `pathPrefix` be added to the URLs in `staticUrls` too?

ameistad:
Yes, that's probably best :+1: I'll do some edits.

ameistad:
Adding `pathPrefix` to every `url` is a little trickier than I initially imagined. There's multuple nested `url` properties in for example `img` and `video`. How about just letting sitemap.js handle the pathPrefix by just adding it to `hostname`?
```js
    const sitemap = require('sitemap').createSitemap({
      hostname: config.siteUrl + pathPrefix,
      cacheTime: options.cacheTime,
      urls: [...generatedUrls, ...staticUrls]
    })
```

hjvedvik:
Yes, that is probably a better approach. You might need to ensure there are no trailing slashes in `siteUrl` and `pathPrefix` before they are joined.

ameistad:
Did some edits and this removes trailing slashes from siteUrl if there are one or more. Do you think it's necessary to also remove from pathPrefix? The example in the docs shows pathPrefix without the slash.

hjvedvik:
The `pathPrefix` is normalized by Gridsome so it should be okay. I haven't been able to test the changes yet, but the code looks good to me :)

PR #412, opened 2019-05-09
chore(filesystem): update docs for index option

xemoka:
Looks like this got missed being documented, found via discord.


hjvedvik:
You're right. Thanks :)

PR #423, opened 2019-05-13
Updated @gridsome/source-drupal README for 0.6.0 changes

matt-e-king:


hjvedvik:
Thanks @matt-e-king Would it be better to use `:title` or `:field_name` in the route examples?

matt-e-king:
@hjvedvik for `'node--article': '/articles/:slug',` I was still in the mind set that this plugin would be responsible for "slugifying" the `title` and adding a `slug` field to each node, worrying that if the route used `title` it would look something like `/article/This Is The Title`.

I now see upon closer inspection, you are already accounting for that :)

For `'taxonomy_term--tags': '/tags/:name'`, I know we are account for both `title` and `name` in the plugin, like so: `title: fields.title || fields.name` but my thinking was Drupal doesn't return a `title` for tags, it just returns `name` by default - so I was trying to keep that relationship clear. (Also, I don't see `field_name` in the node)

That said, I appreciate the consistency you are trying to create across ALL the source plugins, so I will definitely defer to your judgement. I've updated all example routes to `:title` in the latest commit. I've also updated the starter package to reflect that change.

hjvedvik:
You're right :) I forgot that Drupal taxonomies just has `name`.

packages/source-drupal/README.md
```diff
@@ -107,8 +107,8 @@ module.exports = {
       options: {
         baseUrl: 'https://somedrupalsite.pantheonsite.io',
         routes: {
-          'node--article': '/articles/:slug',
-          'taxonomy_term--tags': '/tags/:slug'
+          'node--article': '/articles/:title',
+          'taxonomy_term--tags': '/tags/:title'
```
hjvedvik:
```suggestion
          'taxonomy_term--tags': '/tags/:name'
```

packages/source-drupal/README.md
```diff
@@ -152,8 +152,8 @@ module.exports = {
         baseUrl: 'https://somedrupalsite.pantheon.io',
         exclude: [ ...defaultExcludes, 'user--user' ], // include the defaults
         routes: {
-          'node--article': '/articles/:slug',
-          'taxonomy_term--tags': '/tags/:slug'
+          'node--article': '/articles/:title',
+          'taxonomy_term--tags': '/tags/:title'
```
hjvedvik:
```suggestion
          'taxonomy_term--tags': '/tags/:name'
```

matt-e-king:
👌 Start package updated as well.

hjvedvik:
Thank you @matt-e-king! It looks good now.

PR #424, opened 2019-05-14
Plugin for Vue components in markdown

hjvedvik:
This plugin will make it possible to create pages with Markdown and Vue components. It can be used to replace `src/pages` completely or just for a section on your website.

## Example usage for a documentation section

 ```js
module.exports = {
  plugins: [
    {
      use: '@gridsome/vue-remark',
      options: {
        typeName: 'Post',
        baseDir: './content/blog',
        pathPrefix: '/blog',
        layout: './src/templates/Post.vue'
      }
    }
  ]
}
```

## Example `.md` file

 ```jsx
---
title: A cool title
---
import Message from '~/components/Message.md'
import Button from '~/components/Button.vue'
import labels from '~/data/labels.json'

# {{ $frontmatter.title }}

<Message>Lorem ipsum dolor sit amet...</Message>
<Button @click="send">{{ labels.send }}</Button>

<script>
export default {
  methods: {
    send () {
      // ...
    }
  }
}
</script>
```

## Options

It takes any options that the `filesystem` plugin supports plus the following:

#### layout

 - Type: `string | object`

 Path to the Vue component that will be used as layout for all pages this plugin creates. The option can also be an object with `component` and `props`. Each page can also override this option in their front matter section.

 ```js
layout: './src/layouts/Default.vue'
```
```js
layout: {
  component: './src/layouts/Default.vue',
  props: {
    fullWidth: false
  }
}
```

#### includePaths

- Type: `Array`
- Default: `[]`

Paths or regex that should be parsed by this plugin. Use this option if you want to import `md` files as Vue components. Imported `md` components will not use the `layout` option above.


vpeti:
@hjvedvik awesome work, just what I needed!
I plan to switch from `source-filesystem`, however I also use `refs`. Is it possible to also have that here? 
There seems to be a lot of overlap with  `source-filesystem`, I wonder if this should be just an option.

danrocha:
Looks great.
I assume the .md files only need the `<script>` section if necessary for its components, right?
Meaning, simple .md files without Vue components would also be parsed normally?

hjvedvik:
@vpeti Yes, we'll implement the `refs` option for this plugin too.

I agree that it might look like this plugin overlaps the `filesystem` plugin. But they work very differently internally. This plugin transforms the markdown into Vue components and compiles them with webpack like normal Vue pages. That makes it possible to generate HTML markup and let Vue hydrate it in the browser. I'm not sure how well it will perform for huge blogs but should work well for a docs section or smaller blogs etc. We can, of course, look into if it could be an option for the `filesystem` plugin to do this instead :)

@danrocha Yes, the `script` tag is optional. And you can also register components globally to use them without importing.

geomagilles:
Hello - as the PR takes some time to be validated - is there any way to use this before being merged? Sorry if this question is too obvious.

sanscheese:
Great work on this! Opens up a whole lot of potential. Have just been trying to test it locally and seem to have a quirk.

I've followed the steps in the blog post (so with same setup in `gridsome.config.js`) and created a simple component: `~/components/MyCounter.vue`

In my md file in `./content/docs/tester.md` I have:

```
---
title: Testing page
excerpt: Lorem Ipsum is simply dummy text.
---

import MyCounter from '~/components/MyCounter.vue';

## Here we go

> {{ $frontmatter.excerpt }}

<MyCounter />
````

Which works 🎉 but if I remove the semi-colon the build process locks up with an error 😞:

```
10% building 0/2 modules 2 active ...e=custom&index=0&blockType=static-query(node:22199) UnhandledPromiseRejectionWarning: SyntaxError: Unexpected token, expected ";" (1:50)
    at Parser.raise (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:6322:17)
    at Parser.unexpected (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:7638:16)
    at Parser.semicolon (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:7620:40)
    at Parser.parseImport (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:11111:10)
    at Parser.parseStatementContent (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:9861:27)
    at Parser.parseStatement (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:9763:17)
    at Parser.parseBlockOrModuleBlockBody (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:10340:25)
    at Parser.parseBlockBody (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:10327:10)
    at Parser.parseTopLevel (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:9692:10)
    at Parser.parse (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:11209:17)
    at parse (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/@babel/parser/lib/index.js:11245:38)
    at exports.genImportBlock (/Users/ben/Sites/Open Source/gridsome/gridsome/packages/vue-remark/lib/codegen.js:23:15)
    at compiler (/Users/ben/Sites/Open Source/gridsome/gridsome/packages/vue-remark/lib/toSfc.js:35:17)
    at Function.stringify (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/unified/index.js:344:12)
    at pipelineStringify (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/unified/index.js:41:25)
    at wrapped (/Users/ben/Sites/Open Source/gridsome/gridsome/node_modules/trough/wrap.js:25:19)
(node:22199) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:22199) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
 10% building 1/2 modules 1 active ...ects/testing-md/content/docs/testing.md
```

More on my local setup:
Node version: v11.12.0
Running gridsome repo on branch `feat/plugin-mdvue`

**Comment by hjvedvik** [comment]
Thanks for testing this plugin @sanscheese :) I'll do some debugging and see if I can fix the issue.

**Comment by dcastil** [comment]
This is the final feature which will allow me to migrate from VuePress to Gridsome. Keep it up! 👍

**Comment by perpetualgrimace** [comment]
Looking forward to this feature. I've been rebuilding my portfolio site using Gridsome, but my biggest pain point so far has been trying to keep my existing content structure, with all content (including things like meta descriptions) in markdown files. I think this will help a lot once it's merged in. Nice work! 🙌

**Comment by callumflack** [comment]
@danrocha I see you've been using Vuepress (your FTA site is fantastic & interesting read, BTW). Are you considering Gridsome b/c you're not happy with Vuepress? (Have you tested their [updated blog plugin](https://github.com/vuejs/vuepress/issues/36#issuecomment-498633025) yet?)

**Comment by danrocha** [comment]
@callumflack Thanks!! I haven't updated their blog plugin yet. I like Vuepress, but the development seems slow (some bugs and issues seem to linger a bit too long) and the blog support was an "afterthought". Gridsome has so many good features for blogs built in and the development feels more active. I actually started rebuilding my blog with it a while ago an was loving it until I got stuck with the whole Vue in MD issue (that's why i went with vuepress). As soon as this plugin gets released, I will switch.

**Comment by callumflack** [comment]
Cheers @danrocha. I'll not clog this important thread w/ conversation. Suffice to say I'm also awaiting `.md` blogging support either via Gridsome or Nuxt (Nuxtent _used_ to work ages back…)

**Comment by mittalyashu** [comment]
This PR can make this starter kit possible https://github.com/mittalyashu/gridsome-starter-docs 😅

File: packages/vue-remark/index.js [hunk 1]
```diff
@@ -0,0 +1,165 @@
...
+          component: node.internal.origin,
+          queryVariables: node
+        })
+      }
+    })
+  }
+
+  async parse (source, resourcePath = null) {
+    const { content, ...data } = this.remark.parse(source.trim())
+    const defaultLayout = require.resolve('./src/VueRemarkLayout.js')
```
cco3: It'd be nice if the default layout could be specified in the options.

hjvedvik: The plugin had a `layout` option where you could set a default layout component. But we replaced it with a `component` option instead. They work very differently because the `component` option will be used as the route component and not just wrap the template.

We needed a way to use a component as the route component in order to have a shared `page-query` etc because the layout component could just have a `static-query`. You can read more about the `component` option [here](https://github.com/gridsome/gridsome/blob/75c9ccfa0fd865c205989b0e8fc4ea1777b05be6/packages/vue-remark/README.md#component). Hope it makes sense :)

packages/vue-remark/README.md
```diff
@@ -0,0 +1,151 @@
...
+        baseDir: './src/pages', // required
+      }
+    }
+  ]
+}
+```
+
+## Example component
+
+```jsx
```
cco3: This worked for me with a <style> tag where there were no empty lines, but not one where there were empty lines.

packages/vue-remark/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+```html
+<template>
+  <Layout>
+    <h1>{{ $page.vueRemarkPage.title }}</h1>
+    <VueRemarkContent />
+  </Layout>
+</template>
+
+<page-query>
```
cco3: Is this actually working for templates and page queries?  I'm not able to get it working.  Looking at the code, I suspect it's not since you are only using the createPages api and not doing any of the operations the source-filesystem plugin does, although maybe I'm mistaken about what that accomplishes.

packages/vue-remark/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### baseDir
+
+- Type: `string` *required*
+
+The path to the directory which contains all the `.md` files. A relative path will be resolved from the project root directory.
+
+#### component
+
+- Type: `string`
```
cco3: Would it make sense just to make this a boolean, and then if true, use templates/<typeName>, like source-filesystem?

packages/vue-remark/README.md
hjvedvik: The latest commit should make it work better :)

packages/vue-remark/README.md
cco3: To elaborate, `$page` is null for me, but I can do a static-query just fine.

hjvedvik: I'm not able to reproduce this one. Do you have `component: './src/templates/SomeComponent.vue'`, but `page-query` doesn't work in that component?

cco3: Correct.  (And I'm not aware of any way to get any sort of debugging output regarding the query beyond that.)

cco3: This works:
```
<static-query>
query {
  mytype (path: "/a-const-path") {
    path
    title
  }
}
```

This does not:
```
<page-query>
query {
  mytype (path: "/a-const-path") {
    path
    title
  }
}
```

And of course, using `page-query` with the $id or $path parameters also does not work for me

cco3:
Are the default @gridsome/transformer-remark plugins not enabled?  For example, .md image syntax doesn't appear to be converted to a g-image.

likern:
Am I correct that this plugin allows to use Vue components directly inside Markdown?

Will it allow **to add new Markdown syntax** to be rendered to custom Vue components?
For example, a want to add alert at beginning of each blog post and, instead of mixing Vue and component, I would like to add special Markdown syntax extension using https://github.com/remarkjs/remark which will be rendered to Vue.

packages/vue-remark/README.md
cco3: I figured it out...I need ".vue" at the end of my component name.  Is that expected?

cco3:
If I use the `component` option, should the javascript block in the markdown page still work?

brandonpittman:
@cco3 It should. Is for me.

cco3:
@brandonpittman, my components also don't work using through the component option.  Thinking about it more, I'm not sure why any of this *would* work, if it's chopped up and spat out through a template, you should only get what the template allows to pass through.

packages/vue-remark/README.md
hjvedvik: Good :) Yes, it must include the extension in order to parse the file in Node.js.

hjvedvik:
@cco3 I think we should remove the `content` field in the GraphQL schema to not make any confusion, and only have front-matter fields available. Just have to wait for the schema API to be merged.

@likern Yes, that is what this plugin is meant for :) It uses normal Vue template syntax for the components, so you have to make a custom Remark plugin if you want another syntax.

cco3:
Using updated versions of everything, I can't get this working without setting `chunkName` in the `createPage` calls...no idea if this is a problem with vue-remark or gridsome proper.

lorennorman:
I spent the morning getting this working... _it's awesome!_ 💥 🔥 

I'm sharing my steps because it was not obvious how to do this for me (...and perhaps there's still an easier way?)

1. clone gridsome:
```
git clone git@github.com:gridsome/gridsome.git
```
2. checkout this branch:
```
git checkout feat/vue-remark
```
3. copy the package into your own gridsome project directory (avoids a failure trying to load module `gridsome/graphql`):
```
cp -r gridsome/packages/vue-remark my-project/vue-remark
```
4. use local plugin syntax in `gridsome.config.js`:
```
  ...
  plugins: [
    // Markdown-based Guides
    {
      use: '~/vue-remark',
      options: {
        typeName: 'Guide',
        baseDir: './guides',
        route: '/guides/:slug',
        component: "src/templates/Guide.vue"
      }
    },
  ...
  ]
...
```
5. reference `<VueRemarkContent />` in your template component (if you used `component` option above)
6. read the [correct README](https://github.com/gridsome/gridsome/tree/feat/plugin-mdvue/packages/vue-remark) for further help (not the README-looking blurb at the top of this Pull Request!)

**I do have one lingering issue:** every file loaded this way is getting a Webpack warning: `"webpackChunkName" expected a string, but received: null.`, and indeed, the `.temp/routes.js` shows this:
```
export default [
  {
    path: "/guides/faq",
    component: () => import(/* webpackChunkName: "page--src--templates--guide-vue" */ "/Users/lorennorman/paperize-monorepo/docs/src/templates/Guide.vue"),
    meta: {
      data: true,
      vueRemark: () => import(/* webpackChunkName: null */ "/Users/lorennorman/paperize-monorepo/docs/guides/faq.md")
    }
  },
...
```
...any help with that would be appreciated.

Great work, @hjvedvik!

hjvedvik:
Thanks for testing @lorennorman :) The easiest way to test is to clone the core repo and run `gridsome create test` in the `/projects` folder inside the repo to create a project which can use the local packages. But you must have Yarn installed for it to work.

It seems like this plugin is ready for publishing, then. Just have to fix that chunk name issue and make it work with the changes in Gridsome v0.7.

cco3:
I've noticed that vue-remark does not use the `route` option of `createPage`, so building a large set of pages can take a significant chunk of time (see #548).

hjvedvik:
@cco3 You are right. I wouldn't recommend this plugin for thousands of markdown files with Vue components in them. But it should work well for documentation sites etc. The filesystem source plugin is better suited for a large number of files but doesn't support Vue components.

cco3:
Thank you!  Two questions:
- With the routes change, would you recommend it for thousands of .md files?
- Will this be published on npm sometime soon (or should I just go update my local version of this plugin)?

hjvedvik:
@cco3 The `route` option is currently just a pattern for generating paths, so it will still generate one route per .md file. I want to see if we can get it to work with a single route sometime soon (but every .md file would still need to go through webback).

The plan is to publish this with v0.7 in a couple of hours :)

PR #425, opened 2019-05-14
refactor: remove duplicate exception rethrow

clarkdo:
This `catch` in some files are duplicate because they just rethrow the same exception without any handling.


hjvedvik:
You're right! Thanks :)

PR #427, opened 2019-05-14
refactor: remove unnecessary await on createContentTypes

clarkdo:


hjvedvik:
That `await` keyword was probably added only to align the methods vertically :P

PR #435, opened 2019-05-16
fix: GRIDSOME_MODE in explore

clarkdo:
@hjvedvik I suppose that `serve` is correct mode in `explore`


hjvedvik:
Oh, thanks for noticing! Not sure if `GRIDSOME_MODE` is used in the `explore` command, but good to get things right anyway :)

PR #448, opened 2019-05-21
fixed typo in file

acvander:
There are typos in [this file](https://github.com/gridsome/gridsome/blob/master/gridsome/app/app.js) in the spelling of "graphql". The typos do not currently cause any issues as they are utilizing default imports and the names match throughout the file. The fix passed test on my local machine.


hjvedvik:
Oh, thanks! I should have noticed those before :P

PR #458, opened 2019-05-24
feat(source-airtable): Add airtable source plugin

### LINKED ISSUE #29: Source request: Airtable plugin
Airtable is great for managing sites but Gridsome lacks the support of it for now.

I guess it can be almost the same as the same plugin for Gatsby from https://github.com/jbolda/gatsby-source-airtable

I am going to build this plugin myself. If anybody is doing this already please let us know here. Thanks.

robaxelsen:
Closes #29. Here is the first official draft of the @gridsome/source-airtable plugin. I had a version that worked, but started having issues with error relating to duplicate route names, so could not verify that this is 100% working before pushing:
`TypeName > Failed to add node: Duplicate key for property path: /<routepath>/slug`

Also had a version with class syntax, but with some issues. Decided to push simple object syntax version for now, instead of spending more time on potentially improved class syntax.

@hjvedvik please have a look and advice how to proceed in regards to "duplicate route names" issue, as well as the overall plugin.

### DIFF
--- a/packages/source-airtable/index.js
+++ b/packages/source-airtable/index.js
@@ -2,23 +2,22 @@ const Airtable = require('airtable');
 
 module.exports = function (api, options) {
     const base = new Airtable({apiKey: options.apiKey}).base(options.base);
-    const table = options.table
-    const typeName = table.replace(/ /g, '');
-    const route = typeName.toLowerCase();
+    //const table = options.table
+    //const typeName = options.typeName
+    //const route = options.route
 
     api.loadSource(async store => {
         const contentType = store.addContentType({
-            typeName: typeName,
-            // TODO: Gives me "duplicate route" error
-            route: `/${route}/:slug`
+            typeName: options.typeName,
+            route: `/${options.route}/:id`
         });
 
-        await base(table).select().eachPage((records, fetchNextPage) => {
+        await base(options.table).select().eachPage((records, fetchNextPage) => {
             records.forEach((record) => {
                 const item = record._rawJson;
                 contentType.addNode({
                     id: item.id,
-                    fields: item.fields
+                    ...item.fields
                 });
             });
             fetchNextPage();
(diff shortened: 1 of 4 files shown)


hjvedvik:
Thanks, @robaxelsen :)

You can try to use `:id` in the default route instead. It will make sure every path is unique if no `slug` field exists. And maybe the plugin should have a `route` option to override the default route?

And you can spread `item.fields` in node options. Gridsome v0.6 doesn't need to have custom fields in a `fields` property.

robaxelsen:
Thanks for the help and feedback, @hjvedvik. I made the changes and improvements, and updated the README.md. Please review :smile_cat:

robaxelsen:
@hjvedvik all linting errors fixed, except async nodejs support error. Other source plugins are using the same async pattern, so I guess this error can be ignored?

packages/source-airtable/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "main": "index.js",
+  "keywords": [
+    "gridsome",
+    "gridsome-plugin",
+    "gridsome-source",
+    "airtable"
+  ],
+  "dependencies": {
+    "airtable": "^0.6.0"
+  }
```
hjvedvik:
```suggestion
  },
  "engines": {
    "node": ">=8.3"
  }
```
This will probably make the linter pass.

packages/source-airtable/index.js
```diff
@@ -0,0 +1,23 @@
+const Airtable = require('airtable')
+
+module.exports = function (api, options) {
+  const base = new Airtable({ apiKey: options.apiKey }).base(options.base)
+
+  api.loadSource(async store => {
+    const contentType = store.addContentType({
+      typeName: options.typeName,
+      route: `/${options.route}/:id`
```
hjvedvik:
```suggestion
      route: options.route
```
I think we can make routing optional :)

packages/source-airtable/package.json
robaxelsen: Done.

packages/source-airtable/index.js
robaxelsen: Done.

PR #475, opened 2019-05-30
(Drupal) Add meta from reference to referenced entity

tanc:
As noted in Discord:

When the image fields come through to Gridsome the alt text is lost. The alt text is in a `meta` property but only within the `relationships` property, not on an actual entity.
The JSON looks like this:
```
...
"relationships": {
  "field_image": {
    "data": {
      "type": "file--file",
      "id": "7bd4f306-a450-454d-82af-6395c342240c",
      "meta": {
        "alt": "A screenshot",
        "title": "",
        "width": 800,
        "height": 600
      }
    },
    "links": {
      "self": {
        "href": "http://back.agile.coop/api/media/image/4b2f96cf-0bcb-4d02-aa9d-617041fa1391/relationships/field_image?resourceVersion=id%3A1"
      },
      "related": {
        "href": "http://back.agile.coop/api/media/image/4b2f96cf-0bcb-4d02-aa9d-617041fa1391/field_image?resourceVersion=id%3A1"
      }
    }
  }
}
...
```
When it comes through to Gridsome the meta bit is nowhere to be seen as the relationship is only built with two properties, the type and id.

This pull request adds the meta from a reference to a temporary store and then when the referenced entity is created in GraphQL the stored meta data is added to the fields.

In practical terms this makes the alt and title fields available on the referenced node (file entity).

### DIFF
--- a/packages/source-drupal/gridsome.server.js
+++ b/packages/source-drupal/gridsome.server.js
@@ -94,7 +94,11 @@ class DrupalSource {
       }
     })
 
-    await Promise.all(capturedEntities.map(entity => entity.initialize()))
+    // ensure file entities are processed last
+    const entityList = capturedEntities.slice(0);
+    entityList.sort((a, b) => (a.entityType === 'file--file') ? 1 : 0)
+
+    await Promise.all(entityList.map(entity => entity.initialize()))
   }
 }
 
--- a/packages/source-drupal/gridsome.server.js
+++ b/packages/source-drupal/gridsome.server.js
@@ -24,6 +24,7 @@ class DrupalSource {
     this.defaultExcludes = DEFAULT_EXCLUDES
     this.entities = {}
     this.apiSchema = {}
+    this.storedMeta = {}
 
     api.loadSource(store => this.initialize(store))
   }
@@ -89,7 +90,6 @@ class DrupalSource {
       if (!exclude.includes(entityType)) {
         // creating an instance of the entity class, see ./entities/*
         this.entities[entityType] = new Entity(this, { entityType, url })
-
         capturedEntities.push(this.entities[entityType])
       }
     })
--- a/packages/source-drupal/gridsome.server.js
+++ b/packages/source-drupal/gridsome.server.js
@@ -95,7 +95,7 @@ class DrupalSource {
     })
 
     // ensure file entities are processed last
-    const entityList = capturedEntities.slice(0);
+    const entityList = capturedEntities.slice(0)
     entityList.sort((a, b) => (a.entityType === 'file--file') ? 1 : 0)
 
     await Promise.all(entityList.map(entity => entity.initialize()))
--- a/packages/source-drupal/lib/Entity.js
+++ b/packages/source-drupal/lib/Entity.js
@@ -151,22 +151,21 @@ class Entity {
         result[key] = Array.isArray(data)
           ? data.map(relation => createReference(this.createTypeName(relation.type), relation.id))
           : createReference(this.createTypeName(data.type), data.id)
-      }
-
-      if (data !== null) {
+        // process any meta properties on this relationship and store them
+        // for later use on the related entity
         if (Array.isArray(data)) {
           data.map(relation => {
             if (typeof relation.meta !== 'undefined') {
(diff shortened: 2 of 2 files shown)


tanc:
To avoid the race condition I've sorted the list of content types so that the file type is last. This allows things which reference files to add their meta data. Not ideal but its all I could think of for now. Basically the `media--image` type which references the `file--file` type has to come first as it holds the meta data for `file--file` which we want adding to the file node in Gridsome.

hjvedvik:
Thanks, @tanc :) Are the file entities always called `file--file`? Or are can they be different in Contenta CMS etc?

tanc:
There are other referenced entities that can contain other meta data on the reference, so in theory these should also be placed later in the build process. But the file entity is definitely the one the an media image entity refers to and on that reference the meta data lives, so the file must be created after the media image.

tanc:
I don't think a file entity can get renamed as its a Drupal standard thing

matt-e-king:
@tanc good catch. I had "merging meta data" in the roadmap of the original [pull request](https://github.com/gridsome/gridsome/pull/41) but it got overlooked! Doh!

The way you have it setup now, is it possible that two separate nodes could reference the same entity in their `relationships` object - so the `meta` that is being stored in the `storedMeta` by id will just be overwritten by the second node that is processed? 

Another way to frame this question:
If an Article has a relationship to a File, should the Article own the meta data? Or should the File? Because if another Article can somehow reference that same File, but defines different meta data (alt, height, width), then the two Articles will "compete" for which meta data gets stored in the `storedMeta` object.

My instinct, like yours, is that the File should take ownership of the meta data - but if the above scenario I explained is ever true (for any relationship, not just File) we might have to re-think this approach.

tanc:
> The way you have it setup now, is it possible that two separate nodes could reference the same entity in their relationships object - so the meta that is being stored in the storedMeta by id will just be overwritten by the second node that is processed?

@matt-e-king in theory I think it is possible that other meta could get overridden. The alt text however is unique to the file entity. If this alt text (meta data) was on the referencing field instead I don't think it would be a problem as it could be passed to any image components easily enough from the referencing field, which is basically what I'm doing now:
```
<responsive-image
  stylename="paragraph_image"
  :set="srcset"
  :filename="filename(imageName)"
  :alt="fields.image.field_image.meta.alt"
  fallback="large" />
```
If we had it on the referencing field it would become:
```
<responsive-image
  stylename="paragraph_image"
  :set="srcset"
  :filename="filename(imageName)"
  :alt="fields.image.meta.alt"
  fallback="large" />
```

I did try to add any found meta data to the referencing field but it didn't seem possible, the `createReference` method just takes the type and id as parameters and doesn't allow fields or extra data to be added. But maybe it's possible to add this data later? @hjvedvik maybe you can confirm?

hjvedvik:
@tanc Passing additional data like `meta` to `createReference()` can be difficult. Unless we have two subfields for every reference fields; `node` and `meta`. Maybe it is better if the plugin just adds those fields itself since it's using an API which follows the JSON:API standard? It would be a breaking change but might be worth it. Here is an example query:

```graphql
query {
  drupalArticle(id:"1") {
    field_image {
      node { ... }
      meta { ... }
    }
  }
}
```

tanc:
@hjvedvik yes I think it is best to add the meta to the field in the plugin. The trouble was I couldn't figure out how, which is why I ended up with this workaround to add it to the referenced node. Can you provide an example?

hjvedvik:
@tanc You can try to move [`createReference(this.createTypeName(data.type), data.id)`](https://github.com/gridsome/gridsome/blob/297c587188c25becaa85a7762a069208fe32e680/packages/source-drupal/lib/Entity.js#L152-L153) to another method which looks something like this:

```js
createReference (data) {
  const { createReference } = this.source.store
  const typeName = this.createTypeName(data.type)

  return {
    node: createReference(typeName, data.id),
    meta: data.meta
  }
}
```

matt-e-king:
@hjvedvik @tanc I created another pull request for this based on @hjvedvik recommendation.
https://github.com/gridsome/gridsome/pull/619

Issue #80, opened 2018-12-19
Add graphql-source plugin


jakedohm:
## Summary

Create a plugin that allows a user to connect a remote GraphQL API endpoint, which then gives them full access to the remote schema. 

### Basic example

Gastby Source GraphQL: https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-graphql

### Motivation

- Why are we doing this?: When someone has already put together a GraphQL schema on their DB, it seems fair to assume that Gridsome should be able to use that schema locally, so the user only has to create one schema.
- What use cases does it support?: Anyone who wants to bring in data from a remote GraphQL endpoint.
- What is the expected outcome?: I can install a plugin, pass it a GraphQL endpoint URL, and maybe 1/2 configuration options, and be able to access all of my remote data from within my templates/components.

### Implementation

I'm not 100% sure whether the best implementation is to try to "scrape" the remote schema and use a local version of the schema at build time, or just to use the remote schema directly at build time.

meuschke:
@jakedohm super interested to develop this plugin! I need it for my shopify storefront! 
I tried to merge the remote schema but I am not exactly sure what the return type for addSchemaField is.
This code will not work ...
`    const http = new HttpLink({ uri: `https://${shopName}/api/graphql`, fetch });

    const link = setContext((request, previousContext) => ({
      headers: {
        'X-Shopify-Storefront-Access-Token': `${accessToken}`,
      }
    })).concat(http);

    api.loadSource(async store => {


       const schema = await introspectSchema(link);

       const executableSchema = makeRemoteExecutableSchema({
         schema,
         link,
       });

       const contentType = store.addContentType({
         typeName: 'Shop'
       })
       // TODO -> executableSchema is the wrong type for GraphQLSchema
       contentType.addSchemaField('shop', ({ graphql }) => ({
          type: graphql.GraphQLSchema,
          resolve () {
            return executableSchema
          }
        }))


    })`

Any idea how to merge the schemas correctly ?

Best,
Toni

hjvedvik:
@meuschke Try the [`api.createSchema()`](https://gridsome.org/docs/server-api#apicreateschemafn) hook and see if that works for you :)

jonathanmelville:
Curious if there have been any movement on this.

jakedohm:
@jonathanmelville not that I know of. I'd be happy to test this out and see how it works.

I'll set up a GraphQL server to integrate it with and report back in a few days.

[...]

PR #482, opened 2019-06-03
Feature: Add GraphQL source plugin

jakedohm:
Here is my recommended implementation for the `graphql-source` plugin! It covers all of the bases, I believe: schema stitching, field namespacing, type conflict prevention, and allowing for custom auth with headers!

I'm super excited to see this land in core, and happy to review and updates/changes/nitpicks 😄


jakedohm:
This relates to #80 😄

hjvedvik:
Great work @jakedohm! I don't have much to comment. Just fix the linting and it should be okay :)

hjvedvik:
Do you think `@gridsome/plugin-graphql` would be a better name for this plugin? Just thinking out loud here, since this plugin isn't actually sourcing any nodes and won't work exactly like the other source plugins (thinking about the new routes config etc.). But I don't have anything against `@gridsome/source-graphql` if it's more preferred :)

jakedohm:
I don't have a strong opinion, but I think we should stick with `source-graphql`. It is "sourcing" your data, and that's what Gatsby uses, so it makes sense to follow that convention, IMO.

jakedohm:
@hjvedvik linting errors fixed, should be ready to merge!

packages/source-graphql/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "version": "1.0.0",
```
hjvedvik:
```suggestion
  "version": "0.0.0",
```
We'll probably switch to `1.0.0` when Gridsome reaches v1.0 :)

jakedohm:
@hjvedvik all requested changes have been made!

PR #483, opened 2019-06-04
Fix post folder naming

jojobyte:
When building connected to Ghost Source plugin the posts are put into a folder structure of `dist/Invalid date/Invalid date/Invalid date/whatever-your-slug-was`.

Adding `options.date = options.published_at` fixes that.

### DIFF
--- a/packages/source-ghost/index.js
+++ b/packages/source-ghost/index.js
@@ -109,6 +109,7 @@ class GhostSource {
         options.primary_author = store.createReference(authorTypeName, primary_author.id)
         options.tags = tags.map(tag => store.createReference(tagTypeName, tag.id))
         options.authors = authors.map(author => store.createReference(authorTypeName, author.id))
+        options.date = options.published_at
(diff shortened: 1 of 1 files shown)


hjvedvik:
Hi, which version of Gridsome gives you the `Invalid date` params? Gridsome v0.6 should use the `published_at` field in route params unless there is a bug in core :)

jojobyte:
`"gridsome": "^0.6.3"`

hjvedvik:
Does this happen for all posts or just some of them? I'm only able to reproduce it in the tests if `published_at` is null.

jojobyte:
It happened for all posts. Somehow screwed up my repo last night and haven't been able to reproduce this today. Might have been a weird configuration combination.

hjvedvik:
I'll close this for now instead of adding a `date` field. There was a bug in v0.6.1 which could have caused this. But should work with the latest version :)

Issue #438, opened 2019-05-17
g-image loads twice


markWicht:
## Description

All g-image load twice, max size and correct viewport size

![Screen-2019-05-17_16-16](https://user-images.githubusercontent.com/38689912/57934203-6d56c300-78bf-11e9-88f3-727739e1c514.png)

### Steps to reproduce

Thoroughly tested, should be easy to reproduce, or some platform/node version issue

`gridsome create gridsome-test`  
Index.vue
`<g-image alt="Example image" src="~/assets/pexels-photo-289586.jpeg" width="1920"/>`  
`img {max-width: 100%;}`  

develop and build versions same result

### Expected result

Only load correct size

### Actual result

Loads twice

### Environment

<pre><code>
Libs:
- gridsome version: 0.5.0, 0.6.0, 0.6.1
- @gridsome/cli version: 0.1.0

Browser:
- [ x ] Chrome (desktop) version 74
- [ x ] Firefox version 68
 
For Tooling issues:
- Node version: 11.8
- Platform:  Windows

Others:
Windows 8.1, npm
</code></pre>

dominiquedutra:
I was able to reproduce this on a clean 0.6 gridsome.
Just added an image on ~/images and linked to it using <g-image> component.
Then, I deployed the website to zeit now.

https://tudocondo-gridsome-wordpress.ddutra.now.sh/image

dominiquedutra:
Hey @hjvedvik @markWicht 

I believe this is not related to Gridsome.

I made a sample page to test srcset loading. You can check it out at https://srcset-playground.ddutra.now.sh/

```
<img src="sample-1280x720.jpg"
           srcset=" 
            sample-3840x2160.jpg 3000w,
            sample-1920x1080.jpg 2000w, 
            sample-1280x720.jpg 1000w, 
            sample-640x480.jpg 700w, 
            sample-300x200.jpg 300w" />
```

Using Chrome 74.0.3729.169 64 bits on MacOS and Chrome Device Simulator.

iPad PRO
![image](https://user-images.githubusercontent.com/4117138/58483069-fed2fa00-8135-11e9-9899-5eadc2e616c8.png)

Pixel 2
![image](https://user-images.githubusercontent.com/4117138/58483111-101c0680-8136-11e9-9c0e-22cdde278d4b.png)

iPhone SE
![image](https://user-images.githubusercontent.com/4117138/58483131-1dd18c00-8136-11e9-8ed6-86d694fc2779.png)

[...]

PR #486, opened 2019-06-06
Fix issue #438

dominiquedutra:
Fixed an issue where the g-image component was loading images twice.
Tested on latest browsers (edge, chrome, firefox) simulating HiDPI and iPhone 5 screens and it is working fine - a single image gets loaded on the correct width.
On IE11 the images won't load (only the pixelated SVG) but that also happens on the master branch of the base repo.

### DIFF
--- a/gridsome/app/directives/image.js
+++ b/gridsome/app/directives/image.js
@@ -68,19 +68,9 @@ function loadImage (el) {
     return // src is already switched
   }
 
-  const image = new Image()
-
-  image.onload = function () {
-    el.classList.remove('g-image--loading')
-    el.classList.add('g-image--loaded')
-    
-    el.src = src
-    el.sizes = sizes
-    el.srcset = srcset
-  }
-
-  image.src = src
-
-  el.classList.add('g-image--loading')
-  el.classList.remove('g-image--loaded')
+  el.srcset = srcset
+  el.sizes = sizes
+  el.src = src
+  el.classList.remove('g-image--loading')
+  el.classList.add('g-image--loaded')
 }
--- a/gridsome/app/directives/image.js
+++ b/gridsome/app/directives/image.js
@@ -67,10 +67,13 @@ function loadImage (el) {
   if (!src || el.src.endsWith(src)) {
     return // src is already switched
   }
-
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks for looking into this @dominiquedutra :) Should the classes still be switched in the `onload` event or doesn't it matter much?

gridsome/app/directives/image.js
```diff
@@ -68,19 +68,9 @@ function loadImage (el) {
...
-
-  image.src = src
-
-  el.classList.add('g-image--loading')
-  el.classList.remove('g-image--loaded')
+  el.srcset = srcset
+  el.sizes = sizes
+  el.src = src
+  el.classList.remove('g-image--loading')
+  el.classList.add('g-image--loaded')
```
dcastil: I think it would be better to put this into the onload function. Also don't forget to declare the onload function before you assign the src to the element so the onload function is also called when the image is cached already. 😊

Something like

```js
el.onload = function () {
  el.classList.add('g-image--loaded')
}
el.srcset = srcset
el.src = src
```

dominiquedutra:
Hey, Thanks for the input.

@hjvedvik I did some testing on my code on a image-heavy staging website and it worked just fine.

That being said, I also modified as yours and @dcastil suggestion and it also works flawlessly. You guys probably know better than me so I`ll just commit as you suggestion.

```
function loadImage (el) {
  const src = el.getAttribute('data-src')
  const sizes = el.getAttribute('data-sizes')
  const srcset = el.getAttribute('data-srcset')

  if (!src || el.src.endsWith(src)) {
    return // src is already switched
  }
  
  el.onload = function() {
    el.classList.remove('g-image--loading')
    el.classList.add('g-image--loaded')
  }
  
  el.srcset = srcset
  el.sizes = sizes
  el.src = src
}
```

You can check this in action in here: https://issue-438-reviewed.ddutra.now.sh

Head of master@gridsome/gridsome here: https://tudocondo-staging-no-image-fix.ddutra.now.sh/

hjvedvik:
Seems to work fine. Thanks :)

Issue #488, opened 2019-06-06
Subsequent Gridsome builds causing 404's with JS assets


timcreatewell:
## Description

I'm running into an issue with a Gridsome site in production, which is being hosted on Netlify, however I would imagine this issue would affect most hosting providers. 

Specifically, each time Gridsome builds, new JS assets are generated with new cache-invalidation suffixes.  This unfortunately causes issues for users who are already browsing previous versions of the site.  When they attempt to browse to a new page, the browser complains that the necessary JS is no longer available (404) and the routing does not occur.

### Steps to reproduce

1. Deploy your site to your hosting provider (e.g. Netlify).
2. Visit the site in your browser.
3. Re-deploy your site.
4. Return to your browser and try to browse around your site (without refreshing your browser).
5. Things should fail.

I have also been able to reproduce this locally by running `gridsome build` successively and using a local web server.

### Expected result

The site should gracefully recover, perhaps by doing a page refresh??

### Actual result

The site crashes with multiple 404 errors in the console relating to missing JS files.

### Environment

<pre><code>
Libs:
- @gridsome/cli version: v0.0.9
- gridsome version: v0.6.3
</code></pre>

dominiquedutra:
Browser: Firefox Developer Latest with the console open and preserve log ticked
**Deploy to Zeit Now**
Browsed about 10 pages of the website (not counting the infinite scroll background fetching of json files for next pages)
Rebuild and redeployed over the same project

Browsing pages already visited works
Other pages don`t work - I am getting JSON.parse errors instead of 404s - and the page won't load

![image](https://user-images.githubusercontent.com/4117138/59101567-bff42f80-88ff-11e9-8103-b962829eac78.png)

josepjaume:
I believe a good way to tackle this is to capture the exception and resort to a regular `window.location = url` - that'd refresh the entire app.

hjvedvik:
Yes, I've done some testing and a simple fix would be to use `location.assign(to.fullPath)` in `router.beforeEach` whenever a route component fails to load or a JSON file for a page doesn't exist. We'll try to include a fix next time we publish an update :)

seangwright:
Followed a trail of links on the internet to get here and I'm digging into Gridsome, which seems cool!

Just thought I'd mention that Angular (with which I have a fair amount of experience) solves this very issue using service workers.

[...]

PR #496, opened 2019-06-09
Optionally keep files from old build and don't process images that have already been processed

### LINKED ISSUE #494: Should images be processed on every build?
Marked as a question but its more of a discussion for now.

Do you think images should really be processed on every build?
What about the default behavior be to skip images that have been processed already and just process any new ones?
And with a simple flag on build, we could instruct Gridsome to reprocess everything.

When you have a lot of images being processed by g-image it can take a long time to build and also to deploy if you use some simpler setup like FTP deploy to a web server.

I could give this a try if @tomtev @hjvedvik you agree. And would love to hear everyone elses opinion.

dominiquedutra:
This PR does the following:

1) Expose the build option `--keep={directory1,directory2}` so we can optionally pass directories or files that should be kept from one build to another
2) Don't process images that have already been processed

The benefits are A and B below:

A) In combination, these two modifications allow us to save time on the build by passing the `assets` directory through the keep option.

`gridsome build --keep=assets/static`

With this command, the dist directory is cleared except for the images processed on the last build that were stored on ./dist/assets/static. These images get copied to a temporary directory and then copied back to (outDir) ./dist.
By the time image-processor.js gets to that image, it skips it since it already exists on the directory.
If you want to reprocess everything, just do gridsome build.
If you change any settings on the images being passed to <g-image> or on the graphQL query fetching the image - let's say you change the width - the hash for the file will change and the image will be processed no matter if you pass the `--keep=assets` option. The old image will be kept on ./dist/assets/static too, which leads us to item B.

B) Consider issue #488.

- Your website is live and users are browsing it.
- .html files are not cached by the browser, but users are browsing back and forth through your articles using the back button of the browser or swipe back on mobile - that means they are not reloading the html.
- A new article is published, webhook hits Netlify and your website build process starts.
- New assets get published with different hashes.
- A new user hits your website, he gets the new version, everything works fine.
- Users that were browsing your website during the publish process will get 404 errors when browsing back and forth through pages because the page data does not exist anymore on ./dist directory unless of course, they hit f5.

By allowing us to keep some files from the previous ./dist, we can work around this small issue too.

`gridsome build --keep=assets,data,js,css`

Downside is you will have more files published. Somewhere down the road, you will want to do a full `gridsome build` to clear up the ./dist. That could be ie: scheduled for Sunday at midnight.

B is mostly a consequence of A, because why not? But it kinda offers a solution to issue #488.
Also closes #494.

Looking forward for your input.

### DIFF
--- a/gridsome/index.js
+++ b/gridsome/index.js
@@ -11,8 +11,9 @@ module.exports = ({ context, program }) => {
   program
     .command('build')
     .description('build site for production')
-    .action(() => {
-      wrapCommand(require('./lib/build'))(context, {})
+    .option('-k, --keep <keep>', 'what should be kept from previous build (default: false)')
+    .action(args => {
+      wrapCommand(require('./lib/build'))(context, args)
     })
 
   program
--- a/gridsome/lib/app/loadConfig.js
+++ b/gridsome/lib/app/loadConfig.js
@@ -83,6 +83,7 @@ module.exports = (context, options = {}) => {
   config.filesDir = path.join(config.assetsDir, 'files')
   config.appPath = path.resolve(__dirname, '../../app')
   config.tmpDir = resolve('src/.temp')
+  config.buildTmpDir = resolve('./.buildTmpDir')
   config.cacheDir = resolve('.cache')
   config.dataDir = path.join(config.cacheDir, 'data')
   config.imageCacheDir = resolve('.cache', assetsDir, 'static')
--- a/gridsome/lib/build.js
+++ b/gridsome/lib/build.js
@@ -15,7 +15,27 @@ module.exports = async (context, args) => {
 
   await app.events.dispatch('beforeBuild', { context, config })
 
-  await fs.emptyDir(config.outDir)
+  if (typeof (args.keep) === 'undefined') {
+    await fs.emptyDir(config.outDir) //Default behavior is to clear the outDir completely
+  } else {
+    //Or move files we want to keep to a tmp directory
+    //and later move it back do .dist directory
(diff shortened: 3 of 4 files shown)


hjvedvik:
Thanks for looking into this @dominiquedutra :) This could be a nice feature. But I think it might be better to do this in a plugin instead (or in `gridsome.server.js`). Because the dist folder isn't available during the build process in services like Netlify, etc.

I think something like this could work, but I haven't tested:

1. Move contents of `config.imagesDir` to `config.imageCacheDir` in the `beforeBuild` hook
2. Gridsome checks the cache folder if an image is already processed and copies it to dist
3. Optionally delete `config.imageCacheDir` in the `afterBuild` hook to delete unprocessed images

The issue with missing assets after a new build will be fixed by reloading the page instead :)

dominiquedutra:
Hey @hjvedvik 

Of course, it's better. I did know about the hooks.
I am going to document a basic plugin for your appreciation. Been wanting to try a plugin anyway.

Thanks,

jimmyangel:
Hi @hjvedvik 

In services like Netlify, we would need to add the processed images to the repository. So what about changing your suggestion above with the following?

* Create a directory in the project to keep processed images (```processedImagesDir```); this directory and its content is added to the repository
* In a server plug-in (or ```gridsome.server.js```) do the following:
    * ```beforeBuild```: move content of ```processedImagesDir``` to ```config.imageCacheDir```
    * ```afterBuild```: copy content of ```config.imagesDir``` to ```processedImagesDir```
* In our workflow, it will be required to run ```gridsome build``` before committing and pushing to the repository and triggering a Netlify build

Is this reasonable?

Thanks,

Ricardo

PR #501, opened 2019-06-12
Stabilize image processing worker

u12206050:
Add check for null on g-image component
Added queue to Image Processing Worker to stabilize the processing of large amounts of images.


dominiquedutra:
Hey @u12206050 
Just wondering what you mean by large amounts of images? Do you have a use case?
Thanks,

u12206050:
I have more than 4500 images, which when being processed by Gridsome ends up being close to 11500 images (for various sizes).
In the current state the build just fails. On smaller amounts of images (1000 or so) it might succeed after a few retries. This patch makes it more stable and works for me.

Still need to figure out why image src is `null` during build time, even though it does have a value during development.

dominiquedutra:
That's a lot of images.

@u12206050 whats your opinion on Images being reprocessed every time?
It must be hell for you.
Are you building locally or using Zeit Now / Netlify or similar?

I will be working on a plugin that allows us to save time not processing the same images with same parameters on every build, that's why I ask.

u12206050:
Building both locally and on Netlify. Nice, I thought also of moving the assets image directory to .cache on netlify and then copy that on build back to dist.
Will need a file with hash values to compare if images have changed or even still exist.

dominiquedutra:
@u12206050 AFAIK the processed image hash does not change if you don't change image parameters.

By looking at the image-processor code really quick it seems the hash is built by a combination of all parameters (height, width, fit, quality) so as long as you don't change those, you image final names will be the same every build.

gridsome/lib/build.js
```diff
@@ -148,40 +148,40 @@ async function processFiles (files) {
...
-    info(` 𐄂 - ${timer(hirestime.S)}s`)
     throw err
-  })
+  }
 
   worker.end()
 
-  status(`Processed images (${totalAssets} images)`)
-  info(` ✔︎ - ${timer(hirestime.S)}s`)
+  status(`Process images (${totalAssets} images) - ${timer(hirestime.S)}s\n`)
```
u12206050: The reason there needs to be an `info` after is that the `status` output doesn't add a new line, thus the next info where ever that might be will then write on the same line.

hjvedvik: There is a newline at the end of the string, so it seems to work okay :)

gridsome/app/components/Image.js
```diff
@@ -24,6 +24,11 @@ export default {
     const attrs = data.attrs || {}
     const hook = data.hook || {}
     const res = []
+    
+    if (props.src === null) {
```
hjvedvik: I think you can remove these lines as the source of the bug might have been fixed in https://github.com/gridsome/gridsome/commit/918e76be556aece82ef50c23777f4acc2b2467e8

package.json
```diff
@@ -32,5 +32,8 @@
   "engines": {
     "node": ">=8.3",
     "yarn": "^1.5.1"
+  },
+  "dependencies": {
+    "queue": "^6.0.1"
```
hjvedvik: This dependency isn't used anymore :)

u12206050: Done

PR #504, opened 2019-06-15
refactor: allow pass meta on route

orblazer:
Hello,
I have rename `_meta` to `meta` for allow users to add meta on route.

This can be use for add `requiresAuth`.

Thanks.

### DIFF
--- a/gridsome/lib/pages/__tests__/pages.spec.js
+++ b/gridsome/lib/pages/__tests__/pages.spec.js
@@ -137,6 +137,18 @@ test('create page with custom route', async () => {
   expect(page.internal.isDynamic).toEqual(true)
 })
 
+test('create page with meta', async () => {
+  const { pages: { createPage }} = await createApp()
+
+  const page = createPage({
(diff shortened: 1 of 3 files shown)


hjvedvik:
Thanks for the PR :) I also think adding route meta would be really useful. The reason it is prefixed is that I don't know what's the best way for adding it yet. We also have an undocumented `route` option which will group the pages in a single dynamic route. Using `route` and `meta` together could lead to unwanted results if several pages had the same route but different meta. But we are open for suggestion on how it could be handled properly :)

You can also add `requiresAuth` to the page context and access it as `this.$context.requiresAuth` in your component. But it wouldn't be available in `beforeRouteEnter`...

hjvedvik:
You can add route meta by setting a `route.meta` option when creating pages in Gridsome v0.7:
```js
createPage({
  path: '/hello',
  component: './src/components/Hello.vue',
  route: {
    meta: {
      myMeta: true
    }
  }
})
```

PR #526, opened 2019-06-29
fix load config icon

sprabowo:
bugs: when you add custom config icon below, icon still generate default
```
module.exports = {
  icon: {
    favicon: {
      src: './src/my-favicon.png',
      sizes: [16, 32, 96, 192]
    },
    touchicon: {
      src: './src/my-touchicon.png',
      sizes: [76, 152, 120, 167],
      precomposed: true
    }
  }
}

```
then the pull request can solve it. please review it, thanks @tomtev

### DIFF
--- a/gridsome/lib/app/loadConfig.js
+++ b/gridsome/lib/app/loadConfig.js
@@ -271,14 +271,14 @@ function normalizeIconsConfig (config = {}) {
   res.favicon = typeof icon.favicon === 'string'
     ? { src: icon.favicon, sizes: faviconSizes }
     : Object.assign({}, icon.favicon, {
-      sizes: faviconSizes,
-      src: defaultIcon
+      sizes: icon.favicon.sizes || faviconSizes,
+      src: icon.favicon.src || defaultIcon
     })
 
   res.touchicon = typeof icon.touchicon === 'string'
     ? { src: icon.touchicon, sizes: touchiconSizes, precomposed: false }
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks, @sprabowo :) Your solution will work, but I wonder if it is better to just change to order in `Object.assign()`? The custom values (`icon.favicon` and `icon.touchicon`) should have been the last argument instead of the second.

sprabowo:
> Thanks, @sprabowo :) Your solution will work, but I wonder if it is better to just change to order in `Object.assign()`? The custom values (`icon.favicon` and `icon.touchicon`) should have been the last argument instead of the second.

i just re-order custom values for icon @hjvedvik. i just updated with my second github account :(

hjvedvik:
That's no problem @sprabowo :)

I meant that you can change the arguments passed to `Object.assign()`. Currently, it is `Object.assign({}, custom, default)` but should be `Object.assign({}, default, custom)`.

sprabowo:
> That's no problem @sprabowo :)
> 
> I meant that you can change the arguments passed to `Object.assign()`. Currently, it is `Object.assign({}, custom, default)` but should be `Object.assign({}, default, custom)`.

okay, thanks for the clues @hjvedvik. i've just updated

hjvedvik:
Looks good now @sprabowo :) Thanks!

PR #574, opened 2019-07-23
Permalinks config

hjvedvik:
This PR adds config for paths generated by Gridsome. These are the default values:

```js
module.exports = {
  permalinks: {
    trailingSlash: true,
    slugify: {
      use: '@sindresorhus/slugify',
      options: {}
    }
  }
}
```

Whether this option should be called `permalinks` or not is open for discussion :)

#### Trailing slash

The default value for `trailingSlash: true` will add a slash to all node paths in the GraphQL schema. Static paths passed to `g-link` will not get a slash automatically unless you set `trailingSlash: 'always'`. **Set `trailingSlash: false` to keep the current behavior.**

All paths are without trailing slash internally regardless of the option value.

#### Slugify

Use a custom slugifyer or just change the default options. If you do not want any paths to be slugifyed, set the option to `slugify: false`. The `slugify` option can also be a function:

```js
permalinks: {
  slugify(value) {
    return value
  }
}
```


onecrayon:
Is there an ETA on this PR, or anything third parties can do to help move it toward release? I finally took the time to convert my site to 0.6.x this month, and unfortunately it completely broke the site because I discovered at the last minute that 0.6.x does not respect trailing slashes in route definitions (0.5.x did).

If the PR is still a ways out, are there any recommended short-term hacks for 0.6.8 that I can implement locally that will cause the site to build and use trailing slashes? I need to start publishing again in a week, and I'm worried the site will still be broken.

hjvedvik:
@onecrayon This PR will be included in the 0.7 release. Not sure when we'll publish it but its almost ready now. In the meanwhile, you can include a trailing slash in `g-link` manually:

```html
<g-link to="`${node.path}/`">Read more</g-link>
```

Not an optimal solution, but it will work.

PR #591, opened 2019-08-06
Update Base ID Url

12vanblart:
**Swap from <> to {} so that the url doesn't look like ` https://airtable.com//api/docs#curl/introduction`** 

It looks like the parser is converting `<YOUR_BASE_ID>/api/docs#curl/introduction` to `<your_base_id>/api/docs#curl/introduction</your_base_id>` on https://gridsome.org/plugins/@gridsome/source-airtable

I wasn't able to figure out how to run this locally to double check that this fixes the issue, but if someone could direct me on that, I can add it to the docs.


hjvedvik:
You're right. Thanks for noticing :)

PR #608, opened 2019-08-14
fix g-image sizing for fit inside and fit outside

### LINKED ISSUE #595: Wrong image size when using fit:inside
## Description

When using `fit:inside` on a g-image, the resulting `img` tag has the wrong dimensions applied.

### Steps to reproduce

Obtain an image with dimensions of 600x800 and put it inside the following tag:

`<g-image src="./some_image.png" width="400" height="400" fit="inside"/>`

### Expected result

The `img` tag has the width of the scaled image, which in this case is 300 pixels.

### Actual result

The `img` tag has the width of the bounding box, which is 400 pixels.
As a result, the entire image appears blurred. 

I'm not sure of this is the expected behaviour, so some comments would be appreciated 👍

### Environment

<pre><code>
Libs:
- gridsome version: 0.6.7

Browser:
- all browsers

</code></pre>

fl0cke:
Fixes #595.

### DIFF
--- a/gridsome/lib/app/__tests__/ImageProcessQueue.spec.js
+++ b/gridsome/lib/app/__tests__/ImageProcessQueue.spec.js
@@ -233,6 +233,50 @@ test('do not resize if image is too small', async () => {
   expect(result.srcset).toHaveLength(1)
 })
 
+test('calculate correct image size for fit=inside ', async () => {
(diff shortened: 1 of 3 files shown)


hjvedvik:
Thanks for noticing this and for the PR! Looks good to me :)

PR #667, opened 2019-09-13
create node references for custom fields

baffo:
Adding the functionality to create node references for custom fields that contain a valid postID. It supports referencing either attachments or custom post types. References can be set on any post type, except on attachments (I didn't find a case where this would make sense, but this can be easily expanded).

Requires the `customPostTypeReferences` array to be defined in `gridsome.config.js` under the plugin settings
```
plugins: [{
    use: '@gridsome/source-wordpress',
    options: {
      baseUrl: 'https://example.com', // required
      apiBase: '?rest_route=/',
      typeName: 'WordPress',
      perPage: 100,
      concurrent: 10,
      routes: {
        post: '/:year/:month/:day/:slug',
        post_tag: '/tag/:slug'
      },
      customPostTypeReferences: [
       {
          type: 'attachment',
          sourceField: 'banner_image_id',
          targetField: 'bannerImage'
        },
        {
          type: 'car',
          sourceField: 'car_brand_id',
          targetField: 'carBrand'
        },
        ...
      ]
    }
}]
```

Example of a Query making use of a custom node reference:
```
query {
  allWordPressPost {
    edges {
      node {
        title
        bannerImage {
          id
    	  sourceUrl
          mediaDetails {
             width
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/packages/source-wordpress/index.js
+++ b/packages/source-wordpress/index.js
@@ -124,6 +124,7 @@ class WordPressSource {
   async getPosts (actions) {
     const { createReference } = actions
     const getCollection = actions.getCollection || actions.getContentType
+    const customPostTypeReferences = this.options.customPostTypeReferences
 
     const AUTHOR_TYPE_NAME = this.createTypeName(TYPE_AUTHOR)
     const ATTACHEMENT_TYPE_NAME = this.createTypeName(TYPE_ATTACHEMENT)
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks! The PR looks good. We have a `refs` option for the filesystem source plugin which is doing something similar to this. But our plan is to replace that option with something that can be used across all source plugins. But we are not sure how that configuration should look like yet.

My concern is if we should let each of our source plugins do this in their own way or if we can wait until we find a solution that works for all of them. Any feedback or ideas would be appreciated :)

It is also possible to do it with the new Schema API in Gridsome v0.7:

```js
api.loadSource(({ addSchemaTypes }) => {
  addSchemaTypes(`
    type WordPressPost implements Node @infer {
      bannerImage: WordPressAttachment @proxy(from:"bannerImageId")
      carBrandId: WordPressCar
    }
  `)
})
```

- `@infer` means that Gridsome will add fields that are not defined for the type
- `@proxy` is a helper to use a value from another field on the node

baffo:
I didn't know about the Schema API, I was still at 0.6.7 when writing this change. This looks nice! 

One problem that I keep running into is that wordpress custom fields are strings and the id has to be cast to integer to create the reference. I see this happening with the `@proxy` as well, I didn't dive into the addSchemaTypes syntax yet to see if there's a way this could be overcome. 

I understand your concern. Looking at the bigger picture it makes sense to standardise the way plugins extend this type of functionality. I'm more biased towards the approach that has all project specifics in the config file, so this direction that I took with my PR feels more at home for me.

PR #672, opened 2019-09-16
WordPress Source: Enh(taxonomy): add meta support

edlefebvre:
I suggest adding meta data to graphql for taxonomy terms. 

`add_term_meta`, `update_term_meta` and `get_term_meta` are core WordPress functions. The WP API adds a meta field by default on taxonomy terms.


hjvedvik:
The meta property should be included, thanks for the PR :)

edlefebvre:
You're welcome @hjvedvik!

PR #683, opened 2019-09-19
feat(source-contentful): taxonomy field

mrkstwrt:
Add ability to create a taxonomy type from a specific field on a
contentful content type

For example the following config would create a type called `ContentfulBlogPostCategory` along with `ContentfulBlogPost` which can then be used to build a category page template using that type with belongsTo field in the Graphql schema.

```
{
  options: {
    space: 'xxx',
    accessToken: 'xxx',
    host: 'cdn.contentful.com',
    environment: 'master',
    taxonomy: {
      'Blog Post': 'category'
    }
  }
}
```

I've been using this myself for a while so I thought it might be worth sharing.


hjvedvik:
Thanks for the PR @mrkstwrt and sorry for the late response. The filesystem plugin is doing something similar and other source plugins probably also need this functionality. So I think it would fit better in a separate plugin. But I'm not sure what the plugin should be named or how it should work yet.. :)

PR #687, opened 2019-09-20
Updated the Ghost source plugin based on the new depreciation notices in 0.7.x.

smokeyfro:


hjvedvik:
Can you create another PR for the Airtable changes? We try to keep it separate to generate proper changelogs etc :) Removing the route option will also break sites instead of showing a deprecation warning so this will bump the version up a minor.

smokeyfro:
Sure thing. I reverted the filesystem and airtable related changes. Please take a look and let me know if this is ok. I created separate PR's for the other two.

packages/source-ghost/index.js
```diff
@@ -72,8 +62,7 @@ class GhostSource {
 
   async loadAuthors ({ addContentType, addCollection = addContentType }) {
```
hjvedvik:
```suggestion
  async loadAuthors ({ addCollection }) {
```
Can probably remove the `addContentType` fallback since this version will only work with Gridsome v0.7 anyway :)

packages/source-ghost/index.js
```diff
@@ -82,7 +71,6 @@ class GhostSource {
   async loadPosts ({ addContentType, createReference, addCollection = addContentType }) {
```
hjvedvik:
```suggestion
  async loadPosts ({ createReference, addCollection }) {
```

smokeyfro: PR updated :)

hjvedvik:
Thanks @smokeyfro, looks good now :)

PR #712, opened 2019-09-27
Remove starting slash for some Post Types 'rest_base'

IHIutch:
Looks like some post types (custom post types?) have a "rest_base" starting with /. This makes the fetch request look something like <BASE_URL>/wp-json/wp/v2//api/winners which fails.

Adding trim start to `options.rest_base` should solve the double slash issue

### DIFF
--- a/packages/source-wordpress/index.js
+++ b/packages/source-wordpress/index.js
@@ -98,7 +98,7 @@ class WordPressSource {
         route: this.routes[type]
       })
 
-      this.restBases.taxonomies[type] = options.rest_base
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks! Probably a good idea to remove any leading slash :) Should we do the same with the taxonomy rest bases on [line 101](https://github.com/gridsome/gridsome/blob/10761e71079e18f92b05111fca9a344acbdf43d8/packages/source-wordpress/index.js#L101) too?

PR #716, opened 2019-09-30
Control prefetch/preload the links of assets

igk1972:
Gridsome use vue-server-renderer for build static page. But without settings shouldPrefetch/shouldPreload.
Documentation https://ssr.vuejs.org/guide/build-config.html#client-config
Source https://github.com/vuejs/vue/blob/master/src/server/create-renderer.js#L51

**Problem**

Some sites maybe using many dynamic imports (webpack function). Default generate many-many the link of assets (for prefetch/preload). This is problem for large sites. 
```
  computed: {
    blocks() {
      return this.blocksNames.map(name => {
        const source = name
        return  () => import( /* webpackChunkName: "[request]" */ `~/components/${name}.vue` )
      })
    },
```
Page in dist (before PR)
Before

**Solution**

This PR adding the functionality to control prefetch/preload the links of assets.
Also requires the object 'prefetch' (and/or 'preload') to be defined in gridsome.config.js:
```
  prefetch: {
    mask: '^$', // example - disable all prefetch 
  },
```
Page in dist (after PR)
After

### DIFF
--- a/gridsome/lib/app/loadConfig.js
+++ b/gridsome/lib/app/loadConfig.js
@@ -142,6 +142,9 @@ module.exports = (context, options = {}) => {
 
   config.css = defaultsDeep(localConfig.css || {}, css)
 
+  config.prefetch = localConfig.prefetch || {}
+  config.preload = localConfig.preload || {}
+
   return Object.freeze(config)
 }
 
--- a/gridsome/lib/build.js
+++ b/gridsome/lib/build.js
@@ -74,7 +74,7 @@ async function renderHTML (renderQueue, app, hash) {
   const { createWorker } = require('./workers')
   const timer = hirestime()
   const worker = createWorker('html-writer')
-  const { htmlTemplate, clientManifestPath, serverBundlePath } = app.config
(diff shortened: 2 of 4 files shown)


hjvedvik:
Thanks for the PR @igk1972 :) This would indeed be useful, especially for issues like #762 too. Should the options be grouped in a property called `ssr` or `generator` (or something else)?

Gridsome will eventually have an API where you would be able to add these options as regex or functions inside the worker instead. But it isn't ready yet...

igk1972:
> Should the options be grouped in a property called `ssr` or `generator` (or something else)?

@hjvedvik 
Options can be grouped. But this is not so important.
It is important to have this functionality now. May be undocumented so far. 
People's needs now to make projects with thousands of documents, or a large number of dynamic components.
Your can change code my PR,  like to you needs. 


> issues like #762 too. 

This is the same problem as in this PR.

hjvedvik:
Okay @igk1972. I'll include this, but undocumented for now. It will change in a future version.

Issue #740, opened 2019-10-09
Mailto links in gridsome/vue-remark fail to compile


NickStees:
## Description

Email links In markdown files that are processed with @gridsome/vue-remark cause build failures because I think Gridsome thinks they are relative links?

### Steps to reproduce

Using a `<email@example.com>` or `[mailto](mailto:email2@example2.com)` in markdown files causes `!!assets-loader` errors and the site fails to compile. Here is a [demo repo](https://github.com/NickStees/gridsome-bugs/blob/bug-mailto-in-md/docs/test.md)

### Expected result

Mailto/email links created

### Actual result

<pre><code>ERROR  Failed to compile with 2 errors                                                                                                                                       10:24:00 PM

These dependencies were not found:

* !!assets-loader?to=mailto%3Aemail2%40example2.com&target=_blank&rel=nofollow%20noopener%20noreferrer!mailto:email2@example2.com in ./node_modules/cache-loader/dist/cjs.js?{"cacheDirectory":"node_modules/.cache/gridsome","cacheIdentifier":"7ca0bfbb-vue-loader-template"}!./node_modules/vue-loader/lib/loaders/templateLoader.js??vue-loader-options!./node_modules/vue-loader/lib??ref--17-0!./node_modules/@gridsome/vue-remark/lib/loader.js??ref--17-1!./docs/test.md?vue&type=template&id=2a569395&
* !!assets-loader?to=mailto%3Aemail%40example.com&target=_blank&rel=nofollow%20noopener%20noreferrer!mailto:email@example.com in ./node_modules/cache-loader/dist/cjs.js?{"cacheDirectory":"node_modules/.cache/gridsome","cacheIdentifier":"7ca0bfbb-vue-loader-template"}!./node_modules/vue-loader/lib/loaders/templateLoader.js??vue-loader-options!./node_modules/vue-loader/lib??ref--17-0!./node_modules/@gridsome/vue-remark/lib/loader.js??ref--17-1!./docs/test.md?vue&type=template&id=2a569395&

To install them, you can run: npm install --save !!assets-loader?to=mailto%3Aemail2%40example2.com&target=_blank&rel=nofollow%20noopener%20noreferrer!mailto:email2@example2.com !!assets-loader?to=mailto%3Aemail%40example.com&target=_blank&rel=nofollow%20noopener%20noreferrer!mailto:email@example.com</code></pre>

### Environment

<pre><code>
Libs:
- gridsome version: 0.3.0
[...]

PR #755, opened 2019-10-14
fix: mailto and tel links in g-link

sanscheese:
Have `mailto` and `tel` links also go to the attrs.href of the router-link.

### DIFF
--- a/gridsome/app/components/Link.js
+++ b/gridsome/app/components/Link.js
@@ -1,5 +1,5 @@
 import config from '~/.temp/config.js'
-import { normalizePath } from '../utils/helpers'
+import { normalizePath, isMailtoLink, isTelLink } from '../utils/helpers'
 
 // @vue/component
 export default {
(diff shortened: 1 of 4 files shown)


gridsome/app/components/Link.js
```diff
@@ -83,3 +83,7 @@ function isExternalLink (string) {
 
   return externalRE.test(string)
 }
+
+function isMailToLink (string) { return string.startsWith('mailto:') }
```
hjvedvik:
```suggestion
const isMailToLink = string => string.startsWith('mailto:')
```

gridsome/app/components/Link.js
```diff
@@ -83,3 +83,7 @@ function isExternalLink (string) {
...
+
+function isTelLink (string) { return string.startsWith('tel:') }
```
hjvedvik:
```suggestion
const isTelLink = string => string.startsWith('tel:')
```

sanscheese:
@hjvedvik came across `!!assets-loader` issue when testing with `mailto` links. So included a fix for this too (May cover - #740) 

Not sure if you have any shared utils across `app` and `lib` so i've kept them separate.

hjvedvik:
Yes, `vue-remark` was failing because mailto links were transformed to `g-link`. I made `vue-remark` render them as normal links, but this fix would also fix it. Thanks :)

Issue #292, opened 2019-03-18
g-images with dynamic paths not correctly generating full feature set


bbugh:
### Summary

We have some components that we want to load images in. I've tried many variations on making this work, and in most cases, gridsome simply outputs the literal property path without actually transforming it. Using `require()` will correctly output the path, but then it doesn't output an actual `g-image` feature, just a generic unhashed `img` tag.

I'm quite surprised at this as I can't imagine the tool would be very useful without dynamic images in components.

There is an [existing bug for this](https://github.com/gridsome/gridsome/issues/137) with a proposed solution, however as you'll see below, the proposed solution is actually not working.

### Example code

```vue
<template>
  <div> <!-- ... some other html ... -->

    <!-- FAIL: with computed, it outputs the literal imagePath, broken image -->
    <g-image :src="imagePath" />

    <!-- FAIL: with direct prop, it outputs the literal prop, broken image -->
    <g-image :src="imageSrc" />

    <!-- FAIL: kind of works, but just directly outputs the direct image tag without any g-image features -->
    <g-image :src="webpackImagePath" />

    <!-- FAIL: works, but it's not dynamic! -->
    <g-image src="@/assets/images/literal-image-path.png" /> 
  </div>
</template>

<script>
export default {
  props: {
    imageSrc: {
      type: String,
      required: true
    }
  },
  computed: {
    imagePath () {
      return `@/assets/images/${this.imageSrc}`
    },
    webpackImagePath () {
      return require(`@/assets/images/${this.imageSrc}`)
    }
  }
}
```

### Current behavior

**Literal image path**

Works as expected, but we need dynamic paths.

```
<img src="/assets/static/src/assets/images/literal-image-path.png?width=800" width="800" data-src="/assets/static/src/assets/images/literal-image-path.png?width=800" data-srcset="/assets/static/src/assets/images/literal-image-path.png?width=480 480w, /assets/static/src/assets/images/literal-image-path.png?width=800 800w" data-sizes="(max-width: 800px) 100vw, 800px" class="g-image g-image--lazy g-image--loaded" sizes="(max-width: 800px) 100vw, 800px" srcset="/assets/static/src/assets/images/literal-image-path.png?width=480 480w, /assets/static/src/assets/images/literal-image-path.png?width=800 800w">
```

**[FAIL] Dynamic with property (no require):**

It's outputting the literal image path passed in by the prop or `computed`.

```
<img src="@/assets/images/literal-image-path.png" class="g-image">
```

[...]

PR #758, opened 2019-10-16
Resolve dynamic image paths with g-image

hjvedvik:
This PR attempts to resolve dynamic image paths with `g-image`. A `directory` attribute is required to tell webpack where to find the images. All found images in the directory with an allowed file extension will be processed by default. You can limit extensions to look for with an `extensions="png,jpg"` attribute.

```html
<template>
  <g-image directory="~/assets" :src="imagePath" />
</template>

<script>
export default {
  data: () => ({
    imagePath: 'image.png'
  })
}
</script>
```

Related: https://github.com/gridsome/gridsome/issues/292

### DIFF
--- a/gridsome/app/components/Image.js
+++ b/gridsome/app/components/Image.js
@@ -13,7 +13,10 @@ export default {
     position: { type: String, default: '' },
     background: { type: String, default: '' },
     blur: { type: String, default: '' },
-    immediate: { type: true, default: undefined }
+    immediate: { type: true, default: undefined },
+    directory: { type: String, default: undefined },
+    extensions: { type: String, default: undefined },
+    deep: { type: true, default: undefined }
   },
 
   render: (h, { data, props }) => {
--- a/gridsome/lib/webpack/createBaseConfig.js
+++ b/gridsome/lib/webpack/createBaseConfig.js
@@ -89,7 +89,9 @@ module.exports = (app, { isProd, isServer }) => {
         preserveWhitespace: false,
         modules: [
           require('./modules/html')(),
-          require('./modules/assets')()
+          require('./modules/assets')({
+            defaultImageExtensions: projectConfig.imageExtensions
+          })
         ]
       },
       cacheDirectory,
--- a/gridsome/lib/webpack/modules/assets.js
+++ b/gridsome/lib/webpack/modules/assets.js
@@ -1,37 +1,66 @@
 const isUrl = require('is-url')
 const isRelative = require('is-relative')
 const { isMailtoLink, isTelLink } = require('../../utils')
+const { trimStart } = require('lodash')
 
-module.exports = () => ({
-  postTransformNode (node) {
-    if (node.tag === 'g-link') {
-      transformNodeAttr(node, 'to')
-    }
+const configAttrs = ['directory', 'extensions', 'deep']
 
-    if (node.tag === 'g-image') {
-      transformNodeAttr(node, 'src')
-    }
-  }
-})
+const isStatic = value => /^"([^"]+)?"$/.test(value)
 
-function transformNodeAttr (node, attrName) {
-  if (!Array.isArray(node.attrs)) return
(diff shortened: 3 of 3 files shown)


tomtev:
`directory` could be just `dir`? 
<g-image dir="~/assets" :src="imagePath" />

hacknug:
I don't dislike `dir` but I looked at the name of the props in core's components and almost all of them use full words (eg. `g-image`'s `background`).

regenrek:
I tried this on a fresh build:

* install
clone and install repo
create new project inside `projects` folder
remove node_modules
yarn dev

* add `cat.jpg` to `src/assets`

* add to `index.vue`

```html
<template>
  <Layout>
    <g-image directory="~/assets" :src="imagePath" />
  </Layout>
</template>

<script>
export default {
  data () {
    return {
      imagePath: 'cat.jpg'
    }
  }
}
</script>
```

* error

image


* log

```
attr.value [ { name: 'alt',
    value: '"Example image"',
    dynamic: undefined,
    start: 95,
    end: 114 },
  { name: 'src',
    value: 'require("!!assets-loader?alt=Example%20image&width=135!~/favicon.png")',
    dynamic: undefined,
    start: 115,
    end: 134 },
  { name: 'width',
    value: '"135"',
    dynamic: undefined,
    start: 135,
    end: 146 } ]
genRequireContext

final -> attr.value [ { name: 'directory',
    value: '"~/assets"',
    dynamic: undefined,
    start: 609,
    end: 629 },
  { name: 'src',
    value: 'require.context("!!assets-loader?!~/assets", false, /\\.()$/)("./" + (imagePath || "").replace(/^.\\//, ""))',
    dynamic: false,
    start: 630,
    end: 646 } ]
```

Do I miss something?

tomtev:
@hacknug I agree :)

hjvedvik:
@regenrek Looks like the image extensions are missing from the filter regex. Should work better if you add `extensions="jpg"` to `<g-image>`. I have probably committed some untested changes...

hjvedvik:
@tomtev Yes, there could be a default path in `gridsome.config.js`. It would work slightly different than in `<g-image>` because you wouldn't be able to use webpack aliases. So instead of using `~/assets` you would need to set `./src/assets` as the default path.

@regenrek Your code should work better after the last commit.

Manzurkds:
@hjvedvik Can we do something to make it truely dynamic? Like in this case we have to give a directory attribute and we only have the liberty to have the image name dynamic, but wouldn't it be better if there was a way to make it completely dynamic, which will even be helpful for further implementations like supporting remote images that are placed dynamically in the app?

hjvedvik:
@Manzurkds Yes, it would be very good if we could make it truly dynamic. But the limitation here is how webpack handles assets. webpack will include every image you have in the `/src` folder and process them without the `directory` attribute. Even if you don't use them. Which can be a bit unfortunate. So the `directory` attribute is used to scope it down to one folder.

I think processing remote images should be possible to implement if the URL is a static string. But not if it is a variable. So in some cases, it would probably be better to use an image service like [Cloudinary](https://cloudinary.com/) or [imgix](https://www.imgix.com/) to process remote images on demand.

ramsesgarate:
Is this solved? I can't find anything about this in the documentation, I put directory, but it throws me the following error:

```bash
icons8-html-5-100.png:1 GET http://localhost:8080/about/images/icons/icons8-html-5-100.png 404 (Not Found)
```

I am doing this:

```js
 <g-image 
    class="img-icon"
    directory="~/assets" 
    :src="language.icon"
    width="48"
    height="48"
    :alt="language.text"></g-image>
```

My  version of gridsome is:

```bash
@gridsome/cli v0.3.3
gridsome v0.7.19
```

hacknug:
> Is this solved? I can't find anything about this in the documentation

As you can see on the status of this PR this is just a Draft. So no, it has not been merged.

Issue #213, opened 2019-02-22
Option to change default blur for <g-image>


tomtev:
## Summary

The default blur for small `<g-image>` is 40. This is often too much. There should be an option in **config** to change this globally.

We should probably also change the default so something like 15-20.

hacknug:
@tomtev @hjvedvik should we close this issue? I can open a PR with the default blur changes if that's what's keeping you from closing the issue.

hjvedvik:
Yes, I think so, @hacknug. We have probably just forgotten to close this. The default blur can already be configured with a `images.defaultBlur` option in `gridsome.config.js`. But it's not documented yet because the plan is to move the image processing out to a separate plugin with proper documentation.

PR #760, opened 2019-10-16
config: add images.defaultBlur

sanscheese:
Is this all that's needed for #213 ?

Edit: nope, can see more to be done :)


hjvedvik:
It should be enough :) But I think you just need to add `imageBlurDefault` to `baseconfig` in the tests too. Take a look [here](https://github.com/gridsome/gridsome/blob/master/gridsome/lib/app/__tests__/ImageProcessQueue.spec.js#L9).

sanscheese:
Cheers, had to fix up the way I was trying to add the config value, and the test. Hopefully all good now. :)

sanscheese:
@tomtev & @hjvedvik Looks like some other image config options being considered. It will be better for these to be in an `images` object. i.e.:

```js
module.exports = {
  images: {
    dir: '~/assets/img',
    blurDefault: 10,
  }
  ...
}
```

hjvedvik:
@sanscheese Just noticed that there is actually an [`images`](https://github.com/gridsome/gridsome/blob/master/gridsome/lib/app/loadConfig.js#L110-L114) option today. So it's probably a good idea to have the default blur there too :)

The only option that can be set there is `backgroundColor` which is the default color for the `background` attribute. So maybe the option should be named just `blur` instead? And we could probably also set default `quality` there too.

Just for your info. We are going to try to move the image processor to a separate package as a plugin in v0.8 (haven't started on it yet). So these options will then be deprecated and moved to the plugin options instead when it's published.

tomtev:
@sanscheese That means that the code will still be used. Just moved to a plugin

hjvedvik:
I've seen other libraries often prefix default options with `default*`.  Should we do the same?
```
image: {
  defaultBackground: 'green',
  defaultQuality: 90,
  defaultBlur: 20
}
```
It gets clear that these values can be overridden somewhere else.

sanscheese:
Have updated this to meet feature request for #213 

Can look to refactor config setting conventions later if required.

gridsome/lib/app/loadConfig.js
```diff
@@ -113,6 +113,8 @@ module.exports = (context, options = {}) => {
 
   config.images = { ...localConfig.images }
 
+  config.images.defaultBlur = config.images.defaultBlur || 20
```
hjvedvik: Maybe the default blur should be `40` since that is what it used to be?

sanscheese: @tomtev you have any thought on this? (Based my change on the original request in #213  that 40 is too high for the default).

I don't mind having to change back to 40 if agree.

hjvedvik: You're right, it should be 20 by default :) I'm only concerned about changing the default value in a minor version. But it shouldn't matter too much anyway, or what do you think @tomtev?

sanscheese: @hjvedvik Yeah, that makes sense. I've put it back to `40` to keep less surprises in a minor version. Is easy to change again for next version.

PR #763, opened 2019-10-16
Add join example between taxonomy term and article

btmash:
For a taxonomy term page, instead of getting all the articles and filtering them down, you can use a `Fragment` (or `Inline Fragment`) to join data between two collections. The relationship already exists thanks to JSON API so we can take advantage of that.

### DIFF
--- a/packages/source-drupal/README.md
+++ b/packages/source-drupal/README.md
@@ -245,3 +245,29 @@ Get the details of an individual `DrupalNodeArticle` using `<page-query>` in a G
 ```
 
 Any `relationships` containing a `meta` object in the JSON:API response will be merged as a sibling object alongside `node`. See `field_image` above as an example.
+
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks for the PR @btmash :)

Just a minor comment; can you use `$id` as variable instead and remove the alias?
```diff
- query Tag($path: String!) {
-   tag: drupalTaxonomyTermTags(path: $path) {
+ query Tag($id: ID!) {
+   drupalTaxonomyTermTag(id: $id) {
```
GraphQL can be a bit overwhelming for some, so we try to keep the examples as simple as possible.

And I see that `$id` is `String!` in other query examples. It would be great if you are able to change them to `ID!` too.

btmash:
Thanks for the feedback @hjvedvik! I'd used path as my way to filter by the route I was on and id had been unclear as I was trying to put things together. But I've changed it over to that and added a small note to explore graphql to see what you can filter on and how to figure out what properties you want.

PR #767, opened 2019-10-18
Break out individual Vue-Meta items for use in static rendering

courcelan:
Hi @hjvedvik,

We discussed via Discord a few weeks ago about further breaking out index.html template variables for use when overloading the original index.html. While I personally only needed the `gridsome:hash` broken out, I can see a need for break each out into their own template variable. 
This is my PR to do so.

We have used Gridsome to develop templates for Django (Django-templates) and breaking out these template properties will enable them to be wrapped in Django `{% block %}` statements when trying to integrate.

I look forward to your review and seeing what needs to change if any! Thanks!

---

When overloading (or replacing) the standard index.html file, a user may need to break out the various `${ head }` template variables for use in their chosen platform.

This sanely breaks out the various VueMeta props into their own template variables while retaining the original head template variable that exists currently.

Some rearrangement to the order of the head template variable has been made to group similar functionality/data.

Each new template property has been named to not interfere with current and future naming, while giving the developer implicit understanding of what that template variable exposes. The names of the template variables are up for debate or can change if need be

### DIFF
--- a/gridsome/lib/server/createHTMLRenderer.js
+++ b/gridsome/lib/server/createHTMLRenderer.js
@@ -7,9 +7,19 @@ function createHTMLRenderer (htmlTemplate) {
     return render(Object.assign({
       htmlAttrs: '',
       bodyAttrs: '',
-      scripts: '',
       head: '',
-      app: ''
+      title: '',
+      base: '',
+      hash: '',
+      vueMetaTags: '',
+      vueMetaLinks: '',
+      resourceHints: '',
+      styles: '',
+      vueMetaStyles: '',
+      vueMetaScripts: '',
+      noscript: '',
+      app: '',
+      scripts: ''
     }, variables))
   }
 }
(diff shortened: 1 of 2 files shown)


courcelan:
(whoops looks like I got some weird formatting issues, will upload a fix)

hjvedvik:
Thanks @courcelan :) Looks very good. The new variables must also be listed as empty strings [here](https://github.com/gridsome/gridsome/blob/master/gridsome/lib/server/createHTMLRenderer.js#L8-L12). Because Gridsome only adds the `app` variable during development.

courcelan:
@hjvedvik updated per your request. If and when you merge in for whatever release, I'll go edit the docs with updates on their inclusion.

hjvedvik:
Great @courcelan  :) I'll merge this. But I think we can leave it undocumented for now. Some of the variables might change before v1.0. The `hash` variable, for example, will most likely be removed in v0.8.

courcelan:
@hjvedvik not a problem! Funny about `hash` being removed as that was the whole driver for this as it was causing me problems. That being said, I still think this is completely worthwhile for various platform integrations. so rock on

PR #774, opened 2019-10-23
Add variable disableTitleTemplate

masaakikunsan:
Hi @hjvedvik,
I added disableTitleTemplate to Project configuration.

Even if titleTemplate is currently empty, siteName is entered.
In a real project, you may not want to include siteName or you don't need `-`.
By using disableTitleTemplate, you can edit the title as you like.

### DIFF
--- a/gridsome/lib/app/__tests__/index.spec.js
+++ b/gridsome/lib/app/__tests__/index.spec.js
@@ -59,6 +59,17 @@ test('load env variables by NODE_ENV', () => {
   expect(process.env.PROD_VARIABLE).toEqual('PROD_2')
 })
 
+test('set custom disableTitleTemplate', () => {
+  const config = loadConfig(context, {
+    localConfig: {
+      disableTitleTemplate: true
+    }
+  })
+
+  expect(config.disableTitleTemplate).toEqual(true)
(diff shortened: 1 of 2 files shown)


masaakikunsan:
I'm sorry. It can be solved with `titleTemplate: '% s'`

hjvedvik:
Thanks for the PR @masaakikunsan :) Gridsome currently only checks if the custom `titleTemplate` is a falsy value. But maybe it would be better to check if it is a string instead? And use it regardless of being empty.

masaakikunsan:
@hjvedvik thank you for your reply

Based on this story, I thought that it wasn't good to enter siteName without permission when it was a falsy value, because whether or not `-$ {config.siteName}` is required for each product.

If falsy, how about just `'% s'`?

hjvedvik:
Yes, `%s` should exclude the `siteName` from all title tags be default :) And it can still be modified per page/template. Maybe that is a better approach than adding a new option for disabling it?

masaakikunsan:
@hjvedvik 
Yes, better than adding new options.
If that is OK, close this PR and create a new one.

hjvedvik:
I'll close this then :) We can probably mention `%s` in the docs for those who don't want the site name to be included by default.

PR #778, opened 2019-10-24
Add item.position to collection node

gilesbutler:
Adding item.position will get the sort order from the CMS.

Thanks to @matjack1 from Dato for the hot tip!


hjvedvik:
Thanks, @gilesbutler :)

PR #779, opened 2019-10-25
Adding customEndpoints usage to docs for source-wordpress

jamacon36:
Updating the docs for source-wordpress to reflect `customEndpoints` option.


hjvedvik:
Thanks :)

seffw:
That's all well and good, but I cannot find any reference to using customEndpoints with Gridsome. I can add that to my config, but there seems to not be any documentation on how to use it to display. It's a shame because such features will just get forgotten if people can't use them

PR #806, opened 2019-11-10
feat(page-query): allow src in page-query

sanscheese:


sanscheese:
@hjvedvik Have got the graphql/gql file being read for page-query. But you are right that getting the file watching and page updates working is looking tricky 🤔

Any pointers would be appreciated.

hjvedvik:
Thanks for looking into this @sanscheese :) I was thinking about how to watch the files yesterday and I think adding support for returning a list of file paths in a `dependencies` property is the easiest option. And update the page when any of them is changed. It requires changes in `./lib/pages` so I could try to implement it if its too much work.

And the `src` attribute should also support aliases and extensions from the webpack config. So you could look into the [`enhanced-resolve`](https://www.npmjs.com/package/enhanced-resolve) package. But you have no access to the webpack compiler config now, so I can do some refactoring to make it available.

mklueh:
Thanks for working on this feature. Is there any reason to only do that in <page-query> and not in <static-query> as well?

Can´t <page-query> and <static-query> even be merged into a single query tag or is there a valid reason to have those two separated but doing basically the same?

sanscheese:
@mklueh am assuming you were referring to `<page-query>` and `<static-query>`, but just got stripped out. Currently using files for static queries is working, ie `<static-query src="./some-file.gql" />`.

If this PR eventually handles supporting webpack aliases, then will look to have `static-queries` handle the same (if they don't already).

The `page` and `static` queries are separate blocks as there is more complexity in the `page-query` with things like query variables available to filter based off `id` or `path`.

mklueh:
> @mklueh am assuming you were referring to `<page-query>` and `<static-query>`, but just got stripped out. Currently using files for static queries is working, ie `<static-query src="./some-file.gql" />`.
> 
> If this PR eventually handles supporting webpack aliases, then will look to ensure `static-queries` handle the same (if they don't already).
> 
> The `page` and `static` queries are separate blocks as there is more complexity in the `page-query` with things like query variables available to filter based off `id` or `path`.

Thanks for the hint, you are right, I´ve corrected my post.

Makes sense, thank you.

spacedawwwg:
+1 for this going in

spacedawwwg:
Does this still need extra work? Can I help at all?

hjvedvik:
@spacedawwwg I finished the required changes to make this work a few days ago. And will include them in v0.8. The plan is to also include this PR in the next version :)

spacedawwwg:
I

> @spacedawwwg I finished the required changes to make this work a few days ago. And will include them in v0.8. The plan is to also include this PR in the next version :)

💥 BOOM, @hjvedvik!

thetre97:
@hjvedvik @sanscheese Will this allow multiple `<page-query>` blocks? I.e. I have some fragments that are used across multiple queries, and would be nice to have them in one file that I can import with the query.
Or will imports be available in the `.gql` src file?

hacknug:
> Will this allow multiple `<page-query>` blocks?

Do you really need this? You can already query multiple things on the same block. Here's an example of a multi-query using fragments from one of my sites:

<details>
<summary>show query example</summary>
<pre lang="gql">
query {
  points: allTipsAndTricks(order: ASC, filter: { category: { eq: "Points" } }) {
    ...Parts
  }
  items: allTipsAndTricks(order: ASC, filter: { category: { eq: "Items" } }) {
    ...Parts
  }
  techniques: allTipsAndTricks(order: ASC, filter: { category: { eq: "Techniques" } }) {
    ...Parts
  }
}
fragment Parts on TipsAndTricksConnection {
  edges {
    node {
      id
      path
      name
      itemNo
      description
      attachments {
        ...ImageParts
      }
    }
  }
}
fragment ImageParts on TipsAndTricks_Attachments {
  id
  url
  filename
  size
  type
  thumbnails {
    small {
      url
      width
      height
    }
    large {
      url
      width
      height
    }
    full {
      url
      width
      height
    }
  }
}
</pre>
</details>

thetre97:
But if you want to use that fragment in multiple different page queries, you then have to duplicate it every time. I was thinking you could have another page query block to import single fragments, or even a param in the existing page query block to do that.

hjvedvik:
@thetre97 I think we could look into how to import other GraphQL files in `page-query`. That would make it possible to have fragments in another file. Something like [this](https://www.graphql-tools.com/docs/schema-loading/#using-import-expression), but for queries.

hacknug:
> But if you want to use that fragment in multiple different page queries, you then have to duplicate it every time. I was thinking you could have another page query block to import single fragments, or even a param in the existing page query block to do that.

Ah, yes. Thought you were trying to have multiple queries on the same file and was asking for a way to do that :+1:

hacknug:
@hjvedvik another related thing that would be useful would be a way to use the same fragment with multiple types. One of my sites is fetching content from Airtable and I get multiple types for what could be just one (the issue is Airtable doesn't tell you the type you're dealing with so the plugin creates one per `TypeName`). A workaround would be defining types just once on the plugin and check what kind of content you're dealing with for every column but as soon as they change something, things would break. The current approach feels safer so that's why I didn't propose those changes.

Maybe [Interfaces](https://graphql.org/learn/schema/#interfaces) could help achieve this?

mklueh:
@hjvedvik any news on this feature request?

leightonhughes:
Hi there, just wondering when this will be released? Is there a way to get the feature before the releaase with NPM?
Have just tried to do this but doesn't seem to work. npm install github:<githubname>/<githubrepo>[#<commit-ish>]

noxify:
@leightonhughes something like could work:

```
npm install git://github.com/gridsome/gridsome.git#master --save
```

Source: https://stackoverflow.com/questions/39732397/install-specific-branch-from-github-using-npm
```

Issue #796, opened 2019-11-03
Working with .graphql files instead of <page-query> or <static-query>


mklueh:
With the GraphQL IDE plugin it´s possible to have autocompletion on your GraphQL queries when they are in .graphql files or in a gql tag, however not so when putting your query in <page-query> or <static-query> html elements.

It´s also not possible to share queries between pages or components.

Is it possible to use .graphql or gql tags instead of the html elements with Gridsome?

james-agile-coop:
+1 for this, syntax could be something like:

`<page-query src="~/graphql/query.gql" />`

mklueh:
@james-agile-coop Would be convenient.

ewatch:
Hi together,

just as a feedback to my PR.
I made a first implementation for the above requested feature.
However it still needs some improvement.
Any ideas are welcome.

Thanks and best regards,
Ewatch

hjvedvik:
@ewatch There was another PR (#806) which also allows `src` on `<page-query>` :) I have some suggestions in this [comment](https://github.com/gridsome/gridsome/pull/806#issuecomment-552355309). But it also requires a few improvements for the watcher etc. So I'll take a look at it next week :)

ewatch:
Hey @hjvedvik,

didn't see that. But great, so it can get to an even better version with best of both worlds plus your improvement ideas.
Will also try to support with other additional improvements. 
I think this product is really valuable.

Thanks and best regards,
Jan

PR #812, opened 2019-11-13
Adds logic to load graphql file on the basis of the "source" attribute in the <page-query> tag

ewatch:
#796 

This is a first implementation.
When interpreting the <page-query> custom block in a Vue component it now also checks for an attribute called "source" and if that exists it will load the content of that file and use that as the GraphQL query.

I think there still can be some improvements:
* there might be a better mechanism to get the file content using webpack
* currently when there is a "source" attribute and there would be content for the <page-query> tag the logic would priotize the attribute
* error handling when the file cannot be found
* I would prefer to have an e2e test using that to ensure this functionality 

I had some problems executing the e2e tests.
Is there something I need to be aware of if I run  `npm run test:e2e` ?

### DIFF
--- a/gridsome/lib/plugins/vue-components/index.js
+++ b/gridsome/lib/plugins/vue-components/index.js
@@ -1,48 +1,73 @@
-const path = require('path')
-const compiler = require('vue-template-compiler')
-const { parse } = require('@vue/component-compiler-utils')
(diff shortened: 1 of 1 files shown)


mklueh:
Any news on this pull request? When is it going to be merged?

PR #821, opened 2019-11-17
feat(query): make query available to build

sanscheese:
Just having this as a draft for the moment.

@hjvedvik  I've got reservations making the query source and document available to the prod build by default for performance. Wondering if this should be opt in via a config setting?

Also, thinking if it's available to the static version it should be the same in development.

### DIFF
--- a/gridsome/lib/app/build/executeQueries.js
+++ b/gridsome/lib/app/build/executeQueries.js
@@ -29,8 +29,9 @@ async function executeQueries (renderQueue, { context, pages, schema, graphql },
     invariant(route, `Could not find a route for: ${entry.path}`)
     invariant(page, `Could not find a page for: ${entry.path}`)
 
-    const results = { data: null, context: page.context }
     const document = route.internal.query.document
+    const context = pages._createPageContext(page, entry.queryVariables)
+    const results = { data: null, context }
 
     if (document) {
       if (!validated.has(route.component)) {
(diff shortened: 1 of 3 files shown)


hjvedvik:
Great @sanscheese :) I think we can create a hook to add extra data to the page context to achieve what you want. But it looks like it can be a little tricky to run the same hook in development, so I can take a look at how to do it later.

A hook will let you do something like this in a plugin or `gridsome.server.js`:

```js
const { parse, print } = require('gridsome/graphql')

module.exports = api => {
  api._app.hooks.pageContext.tap('CraftLivePreview', (data, { query, variables }) => {
    const document = parse(query)
    // modify the document
    data.query = print(document)
    data.variables = variables
  })
}
```

*I haven't looked into a good structure for the hooks yet, so they are only available through `api._app.hooks.*`. But will eventually be at `api.hooks.*`.*

sanscheese:
@hjvedvik Nice! Via a hook feels much better, and should make a lot more sense for me to be able to setup a plugin.

(Not being available in development isn't an issue for my needs, was just highlighting if is something that *should* be done)

sanscheese:
@hjvedvik This is working great for me! 😀

Will this be ok for current version or have to wait for the next?

hjvedvik:
I think we can include this in the current version :) It won't affect anyone and the hooks are still not part of the public API yet.

sanscheese:
Awesome. I've been working on a separate plugin for the Craft stuff. Which is pretty much ready for the first release. So nothing more from me on this PR :)

PR #824, opened 2019-11-19
Do not try to read sets array if empty in ImageProcessQueue

mino89:
In ImageProcessQueue I get empty arrays sometime. This error blocks my building process.


hjvedvik:
Thanks, @mino89  :)

PR #839, opened 2019-11-27
Blur svg id hash

igk1972:
This pull request solving the problem difference order the images queue. The assets files are obtained with less changes.

The part is support hosting like Netlify. 

https://www.netlify.com/blog/2017/02/23/better-living-through-caching/

Issue #584, opened 2019-07-31
Different js filenames from one build to another


arttemiuss43:
Every page and template has its own javascript file and names of those files are specific for every build.  Services like Netlify have to reprocess every single generated page every single time because checksum is changed. I have 1700 pages and it takes about five minutes to just process them after Gridsome build is finished. Can we keep JS filenames the same from one build to another?

arttemiuss43:
As I found out Netlify spends about 1 second to process 5 pages. So if the website will have even 20000 pages it will impossible to manage it as the process time will take more than an hour while build time takes less than 2 minutes. I think it is very important when thinking about building large websites. Could you please explain whether Gridsome is able to keep checksum the same if the page is not changed? It is crucial for me. Thank you very much for such an awesome tool you are working on :)

u12206050:
It takes about 4-6 minutes for my current test site which has 7000 pages. However that is after I removed any processing of images. Processing images which were in total around 12000 took an hour and doesn't complete at all on Netlify. So I would suggest moving images to an image cdn and not have Gridsome optimze them, but rather pregenerate a placeholder that you then pass together with the original source to g-image.

igk1972:
See https://github.com/gridsome/gridsome/pull/840

PR #840, opened 2019-11-27
No fingerprints assets - Netlify and cache busting urls

igk1972:
This pull request adding the config option for disable the fingerprints in files. The files are obtained without changes.
**PR no effective without PR https://github.com/gridsome/gridsome/pull/839**

This is very necessary when the site has tens of thousands of HTML pages and as many json files.
Deployment time exceeds half an hour or more.

Netlfy_deploy

For support hosting like Netlify. 
https://www.netlify.com/blog/2017/02/23/better-living-through-caching/

Netlify processes every new file. It takes a lot of time.
Fingerprints are not needed for Netlify. The service runs on information in etag.
Also see https://github.com/gatsbyjs/gatsby/issues/11961#issuecomment-466029938

Probably not everyone needs this feature. But pull request allows now to solve the problem of deploying large sites on Netlify.

Gatsby has the same problems.
https://github.com/gatsbyjs/gatsby/issues/11961


hjvedvik:
Thanks! I can review your PRs tomorrow :)

igk1972:
Related  https://github.com/gridsome/gridsome/issues/584

igk1972:
PR no effective without PR #839

hjvedvik:
Thanks again for a good PR @igk1972 :) I have just one question. Can the option be `urlFingerprints: false` instead? We are trying to have a pattern for options where `false` will disable a default behavior.

hjvedvik:
Or maybe `cacheBusting: false` because the JSON files actually have the hash inside the content instead of the URL. But it might not matter much :)

igk1972:
It is not so important. 

Make (change) an option at your discretion, as you consider it better.

igk1972:
cacheBusting is good :)

hjvedvik:
Okay, then rename it to `cacheBusting` which defaults to `true` for the current behavior :)

igk1972:
I did as needed, added a commit

hjvedvik:
Thank you @igk1972 :)

hacknug:
@hjvedvik any chance we can get something like this working for pages too?

Issue #852, opened 2019-12-03
@gridsome/srouce-contentful support filtering by relationship fields


alexkcollier:
## Description

Currently, when trying to filter Contentful sources by references, you can only filter by the ID of that relationship. This makes filtering by these fields difficult.

### Steps to reproduce

Link to a Contentful space that uses the references content type

### Expected result

Reference fields other than ID would be filterable.

```graphql
query {
  allContentfulPosts(
    filter: { reference: { title: { eq: "Some Title" } } } # Doesn't work
  ) {
    edges {
    node {
      id
      path
      reference {
        id
        title
      }
    }
  }
}
```

### Actual result

Only ID is filterable

```graphql
query {
  allContentfulPosts(
    filter: { reference : { id: { eq: "id-1" } } } # works
  ) {
    edges {
    node {
      id
      path
      reference {
        id
        title
      }
    }
  }
}
```

### Environment

 System:
   - OS: Windows 10
   - CPU: (4) x64 Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz

 Binaries:
   - Node: 12.13.1 - C:\Program Files\nodejs\node.EXE
   - Yarn: 1.19.2 - ~\.yarn\bin\yarn.CMD
   - npm: 6.10.0 - C:\Program Files\nodejs\npm.CMD

 Browsers:
   - Edge: 44.18362.449.0

 npmPackages:
   - gridsome: 0.7.11 => 0.7.11
   - gridsome-plugin-sass-resources-loader: 0.0.3 => 0.0.3

alexkcollier:
#856

PR #856, opened 2019-12-04
fix(source-contentful): allows filtering of posts by reference child fields

alexkcollier:
See #852. This PR allows filtering of Contentful posts by reference fields other than ID.


hjvedvik:
Thanks for the PR @alexkcollier. This change will break inferred reference fields, unfortunately. I think querying by reference fields should be handled in the internal store instead, so it can work for any reference. This is on our todo-list, and I have been doing some experimenting, so I hope to have something finished soon :)

PR #861, opened 2019-12-06
Allow users to pass select options object to Airtable's queries

hacknug:
Airtable requires you to specify a `view` if you want to get your records in the order specified there. This PR adds support for passing this select option object and defaults to an empty one to make it work the same way it did before.


hjvedvik:
Thanks @hacknug :)

PR #865, opened 2019-12-09
Use webpack watchOptions config for webpack-dev-middleware

ostwilkens:
Setting `configureWebpack.devServer.watchOptions` is currently passed to webpack, but actually needs to be set on the middleware to have effect. 
`watchOptions` is needed to enable polling, which is required when running gridsome in a Docker for Windows container.


hjvedvik:
Thanks, @ostwilkens :)

PR #866, opened 2019-12-09
fix(source-contentful): properly handle empty asset fields

### LINKED ISSUE #864: @gridsome/source-contentful: Optional asset fields don't appear in the schema unless they all are not empty
## Description

If I create a Content Type that has an asset field that's optional, it won't appear in the GraphQL schema unless it is given a value on every Content entry.

### Steps to reproduce

1. Create a new Gridsome project and a Contentful project to test with, and configure the @gridsome/source-contentful plugin.
2. Add a new type to the Contentful project that has an asset field that isn't required.
3. Add a Content entry in the Contentful project with an asset in the asset field.
3. Start the Gridsome project and open the GraphQL explorer. Open the "schema" tab.
4. You'll notice that your asset field is listed on your Contentful(TypeName) schema.
5. Add another Content entry with no asset in the asset field.
6. Restart the Gridsome project.

### Expected result

The asset field should still be in the schema for the ContentfulXXX type.

### Actual result

The asset field is no longer in the schema.

### Environment

```
  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
  Binaries:
    Node: 10.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.12.3 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.831.0
  npmPackages:
    gridsome: ^0.7.0 => 0.7.8
```

I'm using version `^0.5.1 => 0.5.1` of @gridsome/source-contentful.

weaversam8:
Iterate based on the fields on the schema, so we can catch the missing fields. Closes #864.

### DIFF
--- a/packages/source-contentful/index.js
+++ b/packages/source-contentful/index.js
@@ -74,7 +74,7 @@ class ContentfulSource {
 
     for (const entry of entries) {
       const typeId = entry.sys.contentType.sys.id
-      const { typeName, displayField } = this.typesIndex[typeId]
+      const { typeName, displayField, fields } = this.typesIndex[typeId]
       const collection = actions.getCollection(typeName)
       const node = {}
 
@@ -84,7 +84,8 @@ class ContentfulSource {
       node.updatedAt = entry.sys.updatedAt
       node.locale = entry.sys.locale
 
-      for (const key in entry.fields) {
+      for (const idx in fields) {
+        const key = fields[idx].id;
         const value = entry.fields[key]
 
         if (Array.isArray(value)) {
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks, @weaversam8. If you want the missing fields to have the correct type in the schema too, then you could look into the [Schema API](https://gridsome.org/docs/schema-api/) to define the schema manually. But it would be best if the Contentful source plugin used the API to generate the schemas for its types.

weaversam8:
> Thanks, @weaversam8. If you want the missing fields to have the correct type in the schema too, then you could look into the [Schema API](https://gridsome.org/docs/schema-api/) to define the schema manually. But it would be best if the Contentful source plugin used the API to generate the schemas for its types.

This is probably a good idea. Created #874 to document this idea, but I probably won't work on it right now.

Issue #904, opened 2019-12-27
Error using `gridsome create project wordpress``


pedrofmiguel:
## Description

Describe the issue that you're seeing.
I created a project with gridsome and when i try to do `gridsome develop` it returns this error
```
TypeError: Cannot destructure property `status` of 'undefined' or 'null'.
    at WordPressSource.fetch (/home/pedroferreira/Documents/Projetos/imo/wp-content/themes/imo-gridsome-zalox/node_modules/@gridsome/source-wordpress/index.js:171:40)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

I didn't even write a line of code, the problem happens during installation 
### Steps to reproduce

1. create project inside theme folder of wordpress using `gridsome create project wordpress`
2. create .env file and create WORDPRESS_URL variable with website url
3. run `gridsome develop` 
4. console error : 

```
TypeError: Cannot destructure property `status` of 'undefined' or 'null'.
    at WordPressSource.fetch (/home/pedroferreira/Documents/Projetos/imo/wp-content/themes/imo-gridsome-zalox/node_modules/@gridsome/source-wordpress/index.js:171:40)
    at process._tickCallback (internal/process/next_tick.js:68:7)

```

Clear steps describing how to reproduce the issue. Please please please link to a demo project if possible, this makes your issue _much_ easier to diagnose (seriously).

### Expected result

Website running in localhost or my website url

### Actual result
```
TypeError: Cannot destructure property `status` of 'undefined' or 'null'.
    at WordPressSource.fetch (/home/pedroferreira/Documents/Projetos/imo/wp-content/themes/imo-gridsome-zalox/node_modules/@gridsome/source-wordpress/index.js:171:40)
    at process._tickCallback (internal/process/next_tick.js:68:7)

```

### Environment

```
 System:
    OS: Linux 4.9 Debian GNU/Linux 9 (stretch) 9 (stretch)
    CPU: (4) x64 Intel(R) Core(TM) i3-3220 CPU @ 3.30GHz
  Binaries:
    Node: 10.18.0 - /usr/bin/node
    npm: 6.13.4 - /usr/bin/npm
  Browsers:
    Chrome: 79.0.3945.88
    Firefox: 68.3.0esr
  npmPackages:
    @gridsome/source-wordpress: ^0.4.0 => 0.4.3 
    gridsome: ^0.7.0 => 0.7.12 
  npmGlobalPackages:
    @gridsome/cli: 0.3.1
```

PR #905, opened 2019-12-27
fix(source-wordpress) Update error formatting in fetch method

thetre97:
Update error formatting in fetch method.

Use `response.status`, instead of trying to destructure `status` from `response.data`, as if there is an error, then data will be null - therefore `status` cannot be destructured from it.


thetre97:
Helps with #904

hjvedvik:
Thanks, @thetre97 :)

PR #923, opened 2020-01-08
use env variable for cpu count and removed unused model from cpus

susanp:
We encountered an issue on circleci regarding the cpu count. They advertise more cpus than you are allowed to use, so we got some memory errors. We solved this by using an environment variable as proposed in this pull request.


hjvedvik:
Thanks @susanp, and sorry for the late reply. The PR looks good to me :)

PR #926, opened 2020-01-09
fix(pages API) Update removePage to return if no page found

thetre97:
Added a return if no page is found, otherwise an error is thrown `Cannot read property 'internal' of undefined`, and the build breaks.

I don't know whether it should just silently return, or perhaps log a message that the page wasn't found?


hjvedvik:
Thanks, @thetre97 I think its okay to just return silently.

PR #927, opened 2020-01-09
feat(Pages API) Add missing find* API methods

thetre97:
Added these missing methods - hope they are implemented correctly...

Can also add documentation for them if needed?


hjvedvik:
Oh, not sure how they got lost.. We will probably start using TypeScript (or JSDoc) to avoid mistakes like this. Thanks again, @thetre97 :)

PR #932, opened 2020-01-10
fix(build): handle errors without an sub-error

oliverpool:
When I added the following performance option:
```
  configureWebpack: {
    // merged with the internal config
    performance: {
      hints: "error",
      maxEntrypointSize: 1
    }
  },
```
The build failed, but the error message was not helpful:
```
(node:12752) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'stack' of undefined
```

This is because the `stats.compilation.errors` does not contain an `.error` subkey in case of a performance issue (when some assets are to big for instance).

This PR fixes this (and uses `flatMap` which looks tidier)

### DIFF
--- a/gridsome/lib/webpack/compileAssets.js
+++ b/gridsome/lib/webpack/compileAssets.js
@@ -25,9 +25,8 @@ module.exports = async (app, defines = {}) => {
 
       if (stats.hasErrors()) {
         const errors = stats.stats
-          .map(stats => stats.compilation.errors)
-          .reduce((acc, errors) => acc.concat(errors), [])
-          .map(err => err.error)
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks, @oliverpool :) I agree that `flatMap` is nicer to use here. My only concern is that it's only supported in Node v11 and above. But v10 will be the active version for a couple of more months. So I think it might be better to use `reduce` for now...

oliverpool:
@hjvedvik thanks for the review, I commented it out :-)

hjvedvik:
Nice @oliverpool. Looks good now :)

PR #951, opened 2020-01-21
allow overriding NODE_ENV during build

miljan-aleksic:
The NODE_ENV could be set to `staging` or  `testing` during the build process. If I am ignoring something, please let me know.

### DIFF
--- a/gridsome/lib/build.js
+++ b/gridsome/lib/build.js
@@ -9,7 +9,7 @@ const { logAllWarnings } = require('./utils/deprecate')
 const { log, info, writeLine } = require('./utils/log')
 
 module.exports = async (context, args) => {
-  process.env.NODE_ENV = 'production'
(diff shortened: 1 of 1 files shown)


hjvedvik:
I'm not sure if setting a custom `NODE_ENV` value is a good idea. Gridsome and plugins often check it for either `development` or `production` to disable/enable behaviors. It might be better to set another env variable, for example, `GRIDSOME_BUILD_ENV`, to check if it's staging, etc. Environment variables prefixed with `GRIDSOME_` will also be available in webpack automatically.

I haven't done much testing yet, so I might be wrong :)

dannon:
My read of this is that it simply respects an existing NODE_ENV if it is set, defaulting to production if not, right?  Proposed code seems like the standard way to handle it -- always forcing 'production' regardless of env seems off.

PR #958, opened 2020-01-29
Update url to baseUrl

smokeyfro:
Fix the url in the "Loading data from undefined" console message while building.


hjvedvik:
Thanks :)

Issue #939, opened 2020-01-15
Router handles normal anchor links as internal URLs


cswebservice:
## Description
We're using a mix of Gridsome pages and Wordpress pages within the same domain.
/:slug will display a gridsome wordpress blog post and /impressum should show a normal page outside gridsome.

The event listener in the following file will handle normal <a> links as internal vue links and show a 404 page instead of making a normal page reload to the given address:
https://github.com/gridsome/gridsome/blob/a4e22d6e8f21be360a7e750dcfea27180910cbe4/gridsome/app/entry.client.js#L39-L78

As you already mentioned as a comment you wanna remove this code. Can you simply add a config option to disable that behavior in general? default value could still be to enable the feature.

### Steps to reproduce
Add a `<g-link to="http://my.domain/foo">` and a `<a href="http://my.domain/bar">`. Both will stay within gridsome.

### Expected result
As in Vue a normal <a> should always redirect like the default behavior of anchors.

### Actual result
Gridsome is listing on clicks and routing pages within the same domain.

### Environment
  System:
    OS: macOS 10.15.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  Binaries:
    Node: 12.13.1 - /usr/local/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.12.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.117
    Firefox: 72.0.1
    Safari: 13.0.4
  npmPackages:
    @gridsome/cli: ^0.3 => 0.3.1 
    @gridsome/source-wordpress: ^0.5 => 0.5.2 
    gridsome: ^0.7 => 0.7.12 
    gridsome-plugin-gtm: ^0.1.1 => 0.1.1

ralph:
I agree with @cswebservice that this event handler breaks external links pointing to the same domain, so this functionality should either be opt-in or go away.

Ideally, Gridsome should expose a new config option `enableAnchorLinks`, which should be `false` by default. So people relying on this functionality can easily opt-in, and others don't get crazy wondering why external links are not working because an event handler has been registered on `document`. Without a config option, it's very hard to get rid of this behaviour, as you don't have a reference to the original event handler. And it breaks stuff in cases when you have a load balancer routing e.g. your marketing pages to gridsome and other pages to a CMS, for example.

a-kriya:
[...]

PR #959, opened 2020-01-29
Add a config flag to disable the event listener on normal anchor links

cswebservice:
See #939

### DIFF
--- a/gridsome/app/entry.client.js
+++ b/gridsome/app/entry.client.js
@@ -43,7 +43,7 @@ document.addEventListener('click', event => {
   const { hostname, port } = document.location
 
   if (
-    config.disableAnchorLinks || // disables this behavior by config settings
+    !config.catchLinks || // disables this behavior by config settings
     event.defaultPrevented || // disables this behavior
     event.which !== 1 || // not a left click
     event.metaKey ||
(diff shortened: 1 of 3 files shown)


hjvedvik:
Thanks, @cswebservice :) We are going to move that global listener to an optional directive instead in the next version of Gridsome. But an option could be a temporary workaround until its ready.

You need to add it [`app/codegen/config.js`](https://github.com/gridsome/gridsome/blob/master/gridsome/lib/app/codegen/config.js) in order to access it in the browser because not all options are included. And we are usually setting options as `false` when disabling a default behavior. So maybe it could be an idea to name the option `catchLinks` instead? And let it be `true` by default.

cswebservice:
Thanks for your feedback! I commited the changes.

Maybe it's a good idea to enable/disable this feature by config in general and also by vue directives for a specific element?

Do you plan to release my changes? Otherwise I'll use my fork.

hjvedvik:
Yes, it could be an idea to just change the default value for `catchLinks` to false in 0.8, and only target the `#app` element if it's enabled. And also have an optional directive to target specific elements instead. It would be useful when used with `v-html`.

The changes you made look good! We will probably publish it one of the next days.

PR #961, opened 2020-01-30
feat(develop): add dev SSL option

morphatic:
Hi there, this is my first PR to Gridsome. I won't be offended at all if this PR needs revisions or otherwise isn't a fit for the project.

This PR adds an optional `localSSL` option to `gridsome.config.js` that allows the local dev server to be configured to use SSL. Here's how it would work:

1. Install [`mkcert`](https://github.com/FiloSottile/mkcert) and setup a locally trusted CA
2. In the root of your project, run `mkcert example.test` where "example.test" is the local domain from which you'd like your dev server to be hosted
3. [Update your `hosts` file](https://linuxize.com/post/how-to-edit-your-hosts-file/) with the line `127.0.0.1 example.test` (again, your desired local domain name)
4. In your project's `gridsome.config.js` file, import the `fs` library and add the following:
  ```js
  const fs = require('fs')

  module.exports = {
    siteName: 'Example',
    plugins: [],
    localSSL: {
      key: fs.readFileSync('example.test-key.pem', 'utf8'),
      cert: fs.readFileSync('example.test.pem', 'utf8')
    }
  }
  ```
5. Run your dev server with `gridsome develop -h example.test`

Now you can visit your app locally at https://example.test:8080

I didn't really know how to write tests for this, so unfortunately, there aren't any. If accepted, I will gladly update the docs to explain how to use the new option.

Thanks for such a great tool for site development!!! 🎉


hjvedvik:
Thanks, @morphatic. Great work! I think local SSL can be useful in many cases. What do you think about also adding `--https`, `--key` and `--cert` as CLI arguments?

And btw, there is a nice package for generating certificates during development called [devcert](https://www.npmjs.com/package/devcert). But it can be implemented sometime later if it's worth it.

morphatic:
@hjvedvik this is GREAT feedback. I hadn't discovered `devcert` until you suggested it. It will streamline a lot of my dev processes.

Here's what changed:

* I added `devcert` as a dependency to `gridsome`
* It's no longer necessary to make any changes to `gridsome.config.js`
* Launching the dev server now looks like: `gridsome develop --https --host example.test` (with the shorter flags this would be `gridsome develop -s -h example.test`)
* The first time you run it, you'll be prompted for your OS password to allow the SSL cert generation

In the limited testing I've done, this "just works" which is pretty amazing. I have only tested on OSX Catalina and High Sierra. "Testing" is not what I'd call robust. There are no unit tests, and in order to test the changes, I manually updated the `node_modules/gridsome` package in an existing website project I'm working on. I don't know if it will perform this smoothly on Windows or Linux.

Some issues:

* The very first time it runs I was getting "unable to open /path/to/local/CA/cert" errors, but it seemed to work anyway
* You get prompted for your password the first time through. This has to be your OS account password, and it would be nice if there were a line break before the prompt and/or some sort of explanation about why you were being prompted.
* Needs testing on Windows and Linux

hjvedvik:
I really like the changes you have made! It probably makes more sense to activate https with just CLI args. Maybe the `devcert.hasCertificateFor(hostname)` can be used to insert a line break if the user needs to type the password?

There are no tests for `gridsome develop` yet... So we can probably create tests for this when creating the development tests some other time.

jason7sc:
Any update on this? This would be an awesome feature to have, especially when developing and testing sites that have packages or external scripts that require SSL to function.

thetre97:
Ditto, could really use this option right about now 😀

samiell (changes requested):
`gridsome/lib/server/Server.js`
.
.
`104    const ssl = useHttps ? await devcert.certificateFor(hostname) : null`
`105   const server = useHttps ? https.createServer(ssl, app) : http.createServer(app)`

fmigliorato:
Hello,
I've an issue using https with cli command, I get this error message: 

> error: unknown option `--https'

fmigliorato:
> @fmigliorato - The option `--https` isn't included in the latest released version ( 0.7.23 ).
> If you want to use it, please use the `master` branch instead of `^0.7.23`

Thank You!

fmigliorato:
> @fmigliorato - The option `--https` isn't included in the latest released version ( 0.7.23 ).
> If you want to use it, please use the `master` branch instead of `^0.7.23`

I get this error while using "master" as the version of gridsome in package.json:

> npm ERR! notarget No matching version found for gridsome@master.
> npm ERR! notarget In most cases you or one of your dependencies are requesting
> npm ERR! notarget a package version that doesn't exist.

You mean version 0.7.2.2?

noxify:
Hey @fmigliorato 

it doesn't work out of the box, since the gridsome repo is a monorepo.

Here my workaround:

1. create a new gridsome project ( `gridsome create testproject` ) 
2. Go to https://gitpkg.vercel.app/ and set `https://github.com/gridsome/gridsome/tree/master/gridsome` as url 
3. copy the npm/yarn command
4. execute the command inside the project
5. Run `gridsome develop --https`

PR #963, opened 2020-01-31
Add tags and author references to static pages for source-ghost plugin

daanbeverdam:
Added tags and author references to Ghost static pages. Previously they were only set for posts.

### DIFF
--- a/packages/source-ghost/index.js
+++ b/packages/source-ghost/index.js
@@ -66,7 +66,7 @@ class GhostSource {
     await this.loadBasicEntity(tags, this.contentAPI.tags)
   }
 
-  async loadPages({ addCollection, createReference }) {
+  async loadPages ({ addCollection, createReference }) {
     console.log(`Loading ${TYPE_PAGE}`)
     const pages = addCollection({
       typeName: this.typeNames.page,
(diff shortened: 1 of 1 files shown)


packages/source-ghost/index.js
```diff
@@ -66,14 +66,45 @@ class GhostSource {
...
+          ? createReference(tagTypeName, primary_tag.id)
+          : null
+        options.primary_author = createReference(authorTypeName, primary_author.id)
+        options.tags = tags.map(tag => createReference(tagTypeName, tag.id))
+        options.authors = authors.map(author => createReference(authorTypeName, author.id))
+
+        pages.addNode(options)
+      })
+
+      if (currentPage === entities.meta.pagination.pages) {
```
zacnomore: I see this is mirroring the loadPosts code below but I wonder about the safety. Could `entities.meta.pagination.pages` become undefined? If so we may be stuck in the while loop unless one of the above operations throws.

daanbeverdam: I haven't tested this thoroughly, but I didn't run into problems with a Ghost install that had only one post (so no pagination). The code could do with some refactoring to keep it DRY though.

hjvedvik:
Thanks, @daanbeverdam and I'm sorry not looking at this sooner... The PR looks good to me. I'll merge it and bump the version to `0.4` :)

PR #985, opened 2020-02-14
Allow GLink and GImage to be used

### LINKED ISSUE #983: Register g-link and g-image in PascalCase
## Summary

I use PascalCase for components name by default in my project, following [official style-guide](https://vuejs.org/v2/style-guide/#Single-file-component-filename-casing-strongly-recommended). 
When I try to use `<GImage>` I get an error: 
`[Vue warn]: Unknown custom element: <GImage> - did you register the component correctly? For recursive components, make sure to provide the "name" option.`

I registered it by myself:
```js
// main.js
import GImage from 'gridsome/app/components/Image';
export default function (Vue, { appOptions, head }) {
  // ...
  Vue.component('GImage', GImage);
  // ...
}
```
But component doesn't work in proper way, I get this in my HTML:
```html
<img alt="Example image" src="@/favicon.png" class="g-image">
```
Props aren't handled.

### Basic example

Not needed.

### Motivation

[Vue component registration docs](https://vuejs.org/v2/guide/components-registration.html#Name-Casing):
> When defining a component with `kebab-case`, you must also use `kebab-case` when referencing its custom element, such as in <my-component-name>.

> When defining a component with `PascalCase`, you can use either case when referencing its custom element. That means both `<my-component-name>` and `<MyComponentName>` are acceptable.

So, please, register custom components in PascalCase by default 🙏

brandonpittman:
This will allow users to use `GLink` as well as `g-link` and `GImage` along with `g-image`.

Fixes #983.

### DIFF
--- a/gridsome/app/app.js
+++ b/gridsome/app/app.js
@@ -16,8 +16,8 @@ import Image from './components/Image'
 import ClientOnly from './components/ClientOnly'
 
 Vue.mixin(graphqlMixin)
-Vue.component('g-link', Link)
-Vue.component('g-image', Image)
+Vue.component('GLink', Link)
+Vue.component('GImage', Image)
 Vue.component('ClientOnly', ClientOnly)
 
 Vue.prototype.$url = url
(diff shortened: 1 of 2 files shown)


hjvedvik:
The PR looks good to me and seems to work well. Thanks, @brandonpittman.

PR #982, opened 2020-02-11
bump up @vue/babel-preset-app to ^4.2.2

clorichel:
Fixes #953.

PR #994, opened 2020-02-18
Add core-js as an explicit dependency

oliverpool:
To prevent issues like https://github.com/oliverfindl/vue-svg-inline-loader/issues/39

According to https://github.com/gridsome/gridsome/pull/982, the `gridsome/app` is expected to somehow depend on `core-js`

### DIFF
--- a/gridsome/package.json
+++ b/gridsome/package.json
@@ -39,6 +39,7 @@
     "color-string": "^1.5.3",
     "columnify": "^1.5.4",
     "connect-history-api-fallback": "^1.6.0",
+    "core-js": "^3.6.4",
     "css-loader": "^2.1.0",
     "didyoumean": "^1.2.1",
     "dotenv": "^6.2.0",


hjvedvik:
Thanks, @oliverpool :) I'm not sure if it will solve the issue because `@babel/preset-env` is expecting `core-js@3` to be in `/node_modules/core-js`, and adding it as a dependency to the `gridsome` package it might end up in `/node_modules/gridsome/node_modules/core-js`, which will not work. I'm not sure how npm or yarn decides which package to hoist, but I'll look into it later today and do some more testing etc.

oliverpool:
Thank you!

As I mentionned in the linked issue, I found a workaround, so you don't have to consider this issue a top-priority (at least for me).

Probably related to https://github.com/babel/babel/issues/10379

oliverpool:
And jest seems to have implemented it https://github.com/xing/hops/pull/969

oliverpool:
This issue seems to only appear when using `yarn`

noxify:
@oliverpool i had this issue also with `npm`.

Berkmann18:
Also getting this error on NPM even with `core-js` set a dev dependency.

zacnomore:
@Berkmann18 You could use the workaround posted here, https://github.com/gridsome/gridsome/issues/1018#issuecomment-597508168

Berkmann18:
@zacnomore That's what I did which works, hopefully, it won't be required.

hjvedvik:
@oliverpool The best solution would probably be if the Vue Babel preset could set an absolute path to its own `core-js` module. But I'll try to include this in the next update of Gridsome. Just have to do some more tests.

zacnomore:
@hjvedvik Could we get a release torn off for this, or back port it to 0.7.13? Without it 0.7.13 needs that workaround to be used.

mediabeastnz:
If you really need `0.7.13` then just add

```json
"core-js": "^3.6.4"
```

To your package.json file and then in the next version you can probably remove it.

zacnomore:
@mediabeastnz see above: https://github.com/gridsome/gridsome/pull/994#issuecomment-601457954
> Also getting this error on NPM even with `core-js` set a dev dependency.

smakinson:
Has this gone out in a release? I'm seeing `Can't resolve 'core-js/modules/es.array.concat'` in gridsome v0.7.18

PR #1003, opened 2020-02-22
Add possibility to ignore files when using vue-remark

### LINKED ISSUE #1002: Exclude paths from vue-remark
`source-filesystem` allows specifying multiple paths so that you can list both included paths and excluded paths.

However, vue-remark simply passes in `'**/*.md'`, so none of that flexibility is retained.  It'd be nice to be able to modify that default `path` value.

Example:
```
options: {
  ...
  path: [
    '**/*.md',
    '!**/README.md',
  ],
}
```

EldoranDev:
Closes  #1002

### DIFF
--- a/packages/vue-remark/README.md
+++ b/packages/vue-remark/README.md
@@ -151,6 +151,8 @@ It is also possible to use slots inside `<VueRemarkContent>`.
 <slot name="tags">
 ```
 
+
+
 #### route
 
 - Type: `string`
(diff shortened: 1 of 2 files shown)


hjvedvik:
Thanks, @EldoranDev! like the ignore based pattern. I'm a little busy with traveling these days, but I can probably do some tests and merge this and a few other fixes for `vue-remark` on Wednesday.

PR #1016, opened 2020-02-28
feat: migrate to leven

### LINKED ISSUE #1001: Migrate to leven to find the closest match for an unknown command
## Summary

Switch over to `leven` for command suggest feature.

### Motivation

As of now [didyoumean.js](https://github.com/dcporter/didyoumean.js/) matching engine is used to suggest matching commands if the user mistypes. But it is not longer actively maintained. Instead I would suggest switching over to [leven](https://github.com/sindresorhus/leven) which is the fastest JS implementation of the Levenshtein distance algorithm.

> `leven` returns the distance between 2 strings which can be made use of to find the exact match.

jamesgeorge007:
closes #1001


hjvedvik:
Thanks, @jamesgeorge007. Looks good to me. There are some minor linter errors. Are you able to fix them before the PR is merged?

PR #1017, opened 2020-02-28
Enable .js and .jsx pages

brandonpittman:
This PR will allow users to use .js and .jsx files as Gridsome pages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 node_modules
 coverage
 dist
+tags
--- a/gridsome/lib/plugins/vue-pages/index.js
+++ b/gridsome/lib/plugins/vue-pages/index.js
@@ -3,15 +3,15 @@ const fs = require('fs-extra')
 const glob = require('globby')
 const slash = require('slash')
 const chokidar = require('chokidar')
-const { trimEnd } = require('lodash')
-const { createPagePath } = require('./lib/utils')
+const {trimEnd} = require('lodash')
+const {createPagePath} = require('./lib/utils')
 
 class VuePages {
-  static defaultOptions () {
+  static defaultOptions() {
     return {}
   }
 
-  constructor (api) {
+  constructor(api) {
     this.api = api
     this.pagesDir = api.config.pagesDir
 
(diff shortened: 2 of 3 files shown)


brandonpittman:
Not getting any love on this one.

hjvedvik:
Thanks, @brandonpittman :) And sorry for not answering you sooner... Supporting pages as `.js` and `.jsx` files would be great! One thing I probably would miss is the ability to have a GraphQL query for the pages. Do you have any plans to also implement that?

brandonpittman:
Didn’t think about that. Only checked that the regular Vue stuff worked. I’ll have to look into how queries need to be setup. Would be grateful if you can point me in the right direction.

hjvedvik:
Vue components are parsed [here](https://github.com/gridsome/gridsome/blob/master/gridsome/lib/plugins/vue-components/index.js#L18-L28). You can look into the [Babel parser](https://babeljs.io/docs/en/babel-parser) and do something similar for .js and .jsx files (maybe .ts and .tsx files too). I'm not sure how the query should be setup, but pages in for example Gatsby export it as a const. And it should be okay to just do the same in Gridsome unless someone has a better idea. We would also need a Babel plugin which removes the exported query when building for production because the it won't be needed later. I hope this will help :)

brandonpittman:
@hjvedvik Would you imagine this sort of API?

```javascript
export default {
	render() { ... },
	pageQuery() { //=> add to `this`}
}
```

Or...

```javascript
import gql from 'graphql'

export default {
	render() {
		let queried = gql`
			query {...}
		`
		return (
			//=> JSX that uses `queried`
		)
	}
}
```

brandonpittman:
@hjvedvik I'm imagining this sort of API in pages.

```javascript
export default {
	render() {…},
	pageQuery() {
		return `
			query {
				…
			}
		`
	}
}
```

brandonpittman:
@hjvedvik After looking at how ugly these components can get in Gatsby, I'm starting to wonder if this is even a good idea. 😂

https://www.gatsbyjs.org/docs/static-query/

PR #1044, opened 2020-03-17
Prefix route with pathPrefix when hard reloading view

### LINKED ISSUE #1043: pathPrefix is not prepended when hard reloading view
## Description

There is an issue in gridsome when a component can't be found, it performs a hard reload. However that hard reload doesn't contain the pathPrefix like it should. 

The problem is, client side your routes don't contain the pathPrefix, which is fine for the client side code, but if those routes can't be found, then it is requested from the server instead, but the server is expecting the route to be prefixed with pathPrefix.

I've noticed this as part of my deployment setup using netlify, where I have it setup to be on a /docs/ sub directory. However, every time I deploy, if I have the site already open, then try to continue navigating the site, the client side routes can't be found and so it requests the page from the server, where I get a 404 because the route it's requesting is without the pathPrefix.

mattbrailsford:
Fixes #1043 

This prefixes the route with the configured `pathPrefix` when attempting to do a full route reload, as client side routes aren't prefixed, but the server will need them.

I'm not sure how to test locally so I'm not sure if the `guard.js` file changes are ok, I just included the config in the same way it was already included in the `entry.client.js` file.

### DIFF
--- a/gridsome/app/entry.client.js
+++ b/gridsome/app/entry.client.js
@@ -28,7 +28,8 @@ if (process.env.NODE_ENV === 'production') {
       .catch(err => {
         // reload page if a component failed to load
         if (err.request && to.path !== window.location.pathname) {
-          window.location.assign(to.fullPath)
+          const fullPathWithPrefix = (config.pathPrefix ?? '') + to.fullPath
+          window.location.assign(fullPathWithPrefix)
         } else {
           next(err)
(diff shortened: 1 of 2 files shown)


hjvedvik:
Thanks, @mattbrailsford. The fix seems good to me. I can do some tests and see if it works.

PR #1047, opened 2020-03-17
feat: pass operationName argument for graphql

### LINKED ISSUE #1045: extend createPages beyond inline operations - add operationName argument for graphql wrapper function
## Summary

api createPages can use only inline operations(queries). Because it has 'graphql function wrapper' as one of the arguments but this wrapper function arguments are not equivalent to one used in express-graphql. Therefore passing 'operationName' is nonexistent thus forcing me to separate graphql operations.

This prevents me from having one `.graphql` file with all operations and fragments.

#### Now we must do
```
api.createPages(async ({ graphql, createPage }) => {
  const context = await graphql(gql_fragment, gql_variables);

  createPage({
    path,
    component,
    context
  });
});
```

#### I would like to do
```
api.createPages(async ({ graphql, createPage }) => {
  const context = await graphql(gql_fragment, gql_variables, gql_operationName);

  createPage({
    path,
    component,
    context
  });
});
```

### Motivation
* it allows basic graphql functionality
* exactly same file as graphiql in http://localhost:8080/__explore
* separate graphql file with operations.

MartinN3:
closes #1045

### DIFF
--- a/gridsome/lib/app/Schema.js
+++ b/gridsome/lib/app/Schema.js
@@ -57,7 +57,7 @@ class Schema {
     return context
   }
 
-  runQuery (docOrQuery, variables = {}) {
+  runQuery (docOrQuery, variables = {}, operationName) {
     const context = this.createContext()
     const func = typeof docOrQuery === 'string' ? graphql : execute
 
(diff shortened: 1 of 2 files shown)


hjvedvik:
Thanks, @MartinN3. Allowing the operation name argument seems like a good idea.

PR #1054, opened 2020-03-19
feat(source-airtable): add option for linked tables

meoweloper:
Updated configuration to accept multiple tables as a source. Introduced new syntax with "tables" array but tried not to introduce breaking changes. It would just show deprecation warning when the old syntax is used.

Added option for linking tables: In Airtable when you set a field to link to another table api just returns  array od ids. Now it maps the actual objects to the ids and returns as a nested resource.

Also added syntax when you link to only one record in the table to conveniently return a first result as an object and not as an array. Inspired by "one to one" convention from relational databases.

I'm already using this implementation on my gridsome project, but I would like to hear comments if I did something wrong or unconventional, and if there are maybe some performance issues with this solution.

Also would like to hear comments about the configuration syntax and naming, I was a little inspired by gatsby source plugin, but I modified it to better fit my personal mental model.

### DIFF
--- a/packages/source-airtable/index.js
+++ b/packages/source-airtable/index.js
@@ -1,28 +1,80 @@
 const Airtable = require('airtable')
+// eslint-disable-next-line node/no-unpublished-require
+const { deprecate } = require('../../gridsome/lib/utils/deprecate')
 
-module.exports = function (api, options) {
-  const base = new Airtable({ apiKey: options.apiKey }).base(options.baseId)
+class AirtableSource {
+  constructor (api, options) {
+    // renamed "baseId" config to "base"
+    options.base = options.base || options.baseId
+    this.base = new Airtable({ apiKey: options.apiKey }).base(options.base)
 
-  api.loadSource(async actions => {
-    const addCollection = actions.addCollection || actions.addContentType
-
-    const collection = addCollection({
-      camelCasedFieldNames: true,
-      typeName: options.typeName,
-      route: options.route
+    api.loadSource(async ({ addCollection, store }) => {
+      this.store = store
+      this.addCollection = addCollection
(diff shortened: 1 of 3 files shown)


msolovyov:
This is the exact issue that I just hit. Linked tables just give back an ID instead of the object. Bravo, hope it gets in.

packages/source-airtable/index.js
```diff
@@ -1,28 +1,80 @@
 const Airtable = require('airtable')
+// eslint-disable-next-line node/no-unpublished-require
+const { deprecate } = require('../../gridsome/lib/utils/deprecate')
```
hjvedvik, line 3:
```suggestion
const { deprecate } = require('gridsome/lib/utils/deprecate')
```
The `gridsome` package is in another location when this plugin is installed in a project. And adding `gridsome` as a peer dependency should get rid of the ESlint message.

meoweloper: Yes, this is the line that I was curious how to solve. Thanks!

meoweloper:
Seems like it didn't solve the lint issue 😄

hjvedvik:
You need to add `gridsome` as a [`peerDependency`](https://docs.npmjs.com/files/package.json#peerdependencies) in the plugin `package.json` :) 

Something like this:
```json
{
  "peerDependencies": {
    "gridsome": "0.x"
  }
}
```

msolovyov:
Would this feature by chance also allow downloading of attached or linked images instead of just the url?

meoweloper:
Thanks @hjvedvik, that did the trick with the linter.

@msolovyov this update doesn't do anything regarding image/file manipulation, I think its fine to show just the url because use case is usually just to embed the image in the html.

Can you give me an example use case why you would need to download it?

PR #1073, opened 2020-04-02
Add suffix test to allow for trailing slash on Pager component

aburkenpas:
This allows for Pager links to have a trailing slash or not have trailing slash depending on vue route


hjvedvik:
Thanks, @aburkenpas. I've tested and you fix seems to work fine :)

PR #1085, opened 2020-04-07
transformer-remark: Populate excerpt with text

onfe:
As mentioned in #237, the `excerpt` field is empty by default. Configuration can be passed to `gray-matter` to populate this field (#284), however `gray-matter` excerpts are not parsed when extracted this way.

This PR addresses this as follows:
- If the user has already got a gray-matter config to populate `excerpt`, it does not overwrite this value.
- If `excerpt` is empty, it populates it with:
  - Raw text, as taken from `sanitizeHTML()`.
  - Newlines are removed and replaced with spaces.
  - The text is limited to at most `length` characters, but will not snip mid-word.
- `excerpt` is added to the schema, including a `length` argument, with a default of 200 characters.

### DIFF
--- a/packages/transformer-remark/index.js
+++ b/packages/transformer-remark/index.js
@@ -122,36 +122,30 @@ class RemarkTransformer {
         args: {
           length: {
             type: GraphQLInt,
-            description: 'Characters',
+            description: 'Maximum length (characters)',
             defaultValue: 200
-          },
-          ellipses: {
-            type: GraphQLString,
-            description: 'String to add to end of excerpt',
-            defaultValue: '…'
           }
         },
-        resolve: async (node, { length, ellipses }) => {
+        resolve: async (node, { length }) => {
           const key = cacheKey(node, 'excerpt')
-          let cached = cache.get(key)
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks, @onfe. Yes, I think it would be nice to generate an except if it doesn't exist :)

Should the `length` argument also limit the `node.excerpt` field? Or would one expect to retrieve the full excerpt since it's not auto-generated? I'm just thinking about a scenario where the `node.excerpt` has content on each node. Then the `length` argument wouldn't do anything.

onfe:
Sure, so I was thinking along the lines that someone set a custom excerpt, you would've already chosen a desired length and it would be confusing if it got snipped.
Alternatively, `length` could always apply but be disabled by setting it to -1, and that's a fairly simple change.

hjvedvik:
Yeah, I agree. Maybe the description for the `length` argument could say "Maximum length for the generated excerpt (characters)" instead to make it clear. Everything else seems good to me.

onfe:
Pulled upstream so we're all up to date, and changed the description based on what you suggested :+1:. Merging this should close #237.

PR #1090, opened 2020-04-08
feat(ghost): Add reading_time for ghost posts

michaelpelletier:
Simple change, just adds the reading_time parameter ([see ghost api docs](https://ghost.org/docs/api/v3/content/#posts))


hjvedvik:
Thanks, @michaelpelletier :)

PR #1096, opened 2020-04-13
Added support for functional components to make use of gql static-query

u12206050:
Exposes `$static` within the render function on the `context` argument at `context.data.$static`

Since functional components don't have state they pair nicely with static queries.

### DIFF
--- a/gridsome/lib/plugins/vue-components/lib/loaders/static-query.js
+++ b/gridsome/lib/plugins/vue-components/lib/loaders/static-query.js
@@ -73,9 +73,17 @@ module.exports = async function (source, map) {
 
       options.__staticData = Vue.observable({ data })
 
-      options.computed = computed({
-        $static: () => options.__staticData.data
-      }, options.computed)
+      if (options.functional) {
+        options.__render = options.render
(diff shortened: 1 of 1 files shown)


hjvedvik:
Great, @u12206050! Would it be better to put `$static` as a prop on `ctx` instead? Since the `data` object is options for `createElement`.

u12206050:
Yeh that is true. Just added it to data since that is where props are also
stored. Ctx should also work.

gridsome/lib/plugins/vue-components/lib/loaders/static-query.js
```diff
@@ -73,9 +73,17 @@ module.exports = async function (source, map) {
 
       options.__staticData = Vue.observable({ data })
 
-      options.computed = computed({
-        $static: () => options.__staticData.data
-      }, options.computed)
+      if (options.functional) {
+        options.__render = options.render
+        options.render = function(h, ctx) {
+          ctx.data.$static = options.__staticData.data
```
hjvedvik, line 79:
```suggestion
          ctx.$static = options.__staticData.data
```

Issue #750, opened 2019-10-11
Incompatible with sass-loader@8


wyattis:
## Description

sass-loader v8 is incompatible with Gridsome's default options. Need to change default options to work with v8 or update the docs to have them install sass-loader@7.* instead. I'm not sure if the v8 options are backwards compatible with v7.

### Steps to reproduce

1. `gridsome create sass-testing`
1. `cd sass-testing`
1. `npm i -D sass-loader node-sass`
1. Change the style tag in *layouts/Default.vue* from `<style>` to `<style lang="sass">` and add some sass code:
```
html, body
    background: red
```
1. `gridsome develop`


### Expected result

Should compile successfully.

### Actual result

```
 ERROR  Failed to compile with 1 errors                                                                                                                                                                                                                       11:40:14 AM

 error  in ./src/transitions/SlideTransition.vue?vue&type=style&index=0&id=7479ebac&lang=sass&scoped=true&

Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
ValidationError: Invalid options object. Sass Loader has been initialised using an options object that does not match the API schema.
 - options has an unknown property 'indentedSyntax'. These properties are valid:
   object { implementation?, sassOptions?, prependData?, sourceMap?, webpackImporter? }
    at validate (C:\Users\wi27\Documents\code\breadboard\docs\node_modules\sass-loader\node_modules\schema-utils\dist\validate.js:50:11)
    at Object.loader (C:\Users\wi27\Documents\code\breadboard\docs\node_modules\sass-loader\dist\index.js:36:28)

 @ ./node_modules/vue-style-loader!./node_modules/css-loader/dist/cjs.js??ref--5-oneOf-1-1!./node_modules/vue-loader/lib/loaders/stylePostLoader.js!./node_modules/postcss-loader/src??ref--5-oneOf-1-2!./node_modules/sass-loader/dist/cjs.js??ref--5-oneOf-1-3!./node_m
odules/cache-loader/dist/cjs.js??ref--0-0!./node_modules/vue-loader/lib??vue-loader-options!./src/transitions/SlideTransition.vue?vue&type=style&index=0&id=7479ebac&lang=sass&scoped=true& 4:14-470 14:3-18:5 15:22-478
 @ ./src/transitions/SlideTransition.vue?vue&type=style&index=0&id=7479ebac&lang=sass&scoped=true&
```

### Environment

<pre><code>
Libs:
- gridsome version: 0.7.7
[...]

PR #1108, opened 2020-04-19
fix(sass-loader): improve default options (#750)

lulzneko:
Change to be compatible with sass-loader@8 as reported in #750.

The option `indentedSyntax` of the sass-loader is now defined under `sassOptions`.
This pull request modifies the default definition of gridsome to be compatible with the sass-loader configuration.

A description of the sassOptions of the sass-loader is given below:
webpack-contrib/sass-loader: Compiles Sass to CSS
https://github.com/webpack-contrib/sass-loader#sassoptions.

### DIFF
--- a/gridsome/lib/app/loadConfig.js
+++ b/gridsome/lib/app/loadConfig.js
@@ -34,7 +34,9 @@ module.exports = (context, options = {}) => {
     split: false,
     loaderOptions: {
       sass: {
-        indentedSyntax: true
+        sassOptions: {
+          indentedSyntax: true
+        }
       },
       stylus: {
         preferPathResolver: 'webpack'


hjvedvik:
Thanks, @lulzneko. Your PR looks good. But could we try to detect which version of `sass-loader` is installed? And use different defaults for v7 and below. It's just so that we don't break projects already using v7 in a minor release :) Or else we can also wait a bit and merge this in when we're about to release Gridsome 0.8 too.

fmakdemir:
How about removing `indentedSyntax` option? `sass-loader` defaults are using it for `.sass` extensions and not using it for `.scss`

I couldn't think of a clean way to detect `sass-loader` version since it might or might not be in use and doing a package existing + version check in `loadConfig.js` didn't seem very nice.

Also, would changing the default config break older ones using `sass-loader@^7.x.x` ?

PR #1127, opened 2020-04-26
fix: allow GET request with query param for the local GraphQL server

sanscheese:
Allowing access the local GraphQL server with GET requests, with the `query` param e.g. `http://localhost:8080/___graphql?query={metadata{siteDescription}}`


sanscheese:
Ahh, looks like it broke a test. So likely have to make sure it's the Graphql endpoint. Is too late here, so will aim to fix it up tomorrow :)

sanscheese:
I've updated this PR to follow closer to the current code.

The unit test looks like a false alarm. See https://github.com/gridsome/gridsome/pull/1129

sanscheese:
Merged in master to confirm tests are holding up

PR #1128, opened 2020-04-27
Do not count remark-cli as a plugin in vue-remark

cco3:
remark-cli cannot be imported/required, so vue-remark should not try
to import it as though it is a plugin.


hjvedvik:
Yes, I agree. That package should be skipped. Thanks, @cco3!

PR #1129, opened 2020-04-27
fix: unit test breaking on escape character in regex

sanscheese:
The unit test have started failing on CircleCI, but am having to fix it with a useless escape character? [^**\\**/]

```
Expected: /^\/page\/([^/]+?)(?:\/(\d+))?(?:\/)?$/i
Received: /^\/page\/([^\/]+?)(?:\/(\d+))?(?:\/)?$/i
```


hjvedvik:
I think the tests began to fail because the Docker image which is used to run the tests recently started to use Node.js v14. And the same tests failed for me too when using that version locally. Your fix seems okay to me :) Thanks, @sanscheese!

Issue #1106, opened 2020-04-18
Add `help` command


AndrewKvalheim:
## Summary

This is confusing:

```console
$ gridsome help

  Please run yarn to install dependencies first.

```

I futilely chased around nonexistent package issues for a while before figuring out this is Gridsome's way of saying "command not found".

### Basic example

```console
$ gridsome help

  Run gridsome --help for usage information.

```

### Motivation

Enhance discoverability.

PR #1130, opened 2020-04-27
fix(cli): help for unknown commands

sanscheese:
In response to https://github.com/gridsome/gridsome/issues/1106


packages/cli/bin/gridsome.js
```diff
@@ -59,8 +59,9 @@ program.arguments('<command>').action(async command => {
     const useYarn = await hasYarn()
 
     console.log()
-    console.log(`  Please run ${chalk.cyan(useYarn ? 'yarn' : 'npm install')} to install dependencies first.`)
+    console.log(chalk.red(`Unknown command ${chalk.bold(command)}`))
```
jamesgeorge007, line 62: Why not taking it outside the `if/else` block?

jamesgeorge007:
```js
program.arguments('<command>').action(async command => {
  const { isGridsomeProject, hasYarn } = require('../lib/utils')
  const availableCommands = program.commands.map(cmd => cmd._name)
  const suggestion = availableCommands.find(cmd => {
    const steps = leven(cmd, command)
    return steps < 3
  })

  console.log(chalk.red(`Unknown command ${chalk.bold(command)}`))

  if (isGridsomeProject(pkgPath) && !suggestion) {
    const useYarn = await hasYarn()

    console.log()
    console.log()
    program.outputHelp()
  } else if (suggestion) {
      console.log()
      console.log(`Did you mean ${suggestion}?`)
  }
})
```

hjvedvik: Yes, that seems like a better solution @jamesgeorge007 :)

sanscheese: @jamesgeorge007 Cheers for that. Agree. Btw, is the change to `suggestion` intentional?

```js
const suggestion = didYouMean(command, availableCommands)
```

```js
const suggestion = availableCommands.find(cmd => {
    const steps = leven(cmd, command)
    return steps < 3
  })
```

jamesgeorge007: #1016

jamesgeorge007: #1001

PR #1140, opened 2020-05-01
Update sharp dependency to ^0.25.2

marcvangend:
When I tried building my gridsome project on a node:14-alpine docker container (a lightweight linux container with Node.js 14), it failed because of the dependency on sharp ^0.23.2. As per the response of the Sharp maintainer in https://github.com/lovell/sharp/issues/2190, this would be solved by upgrading to sharp ^0.25.2 because its prebuilt binaries are no longer geared towards a specific Node.js version.

My own project seemed to run fine on sharp 0.25.2, but I must mention that I did not do extensive testing. 

This is my first gridsome pull request so please let me know if I'm not following the correct procedure here.

### DIFF
--- a/gridsome/package.json
+++ b/gridsome/package.json
@@ -81,7 +81,7 @@
     "portfinder": "^1.0.20",
     "postcss-loader": "^3.0.0",
     "probe-image-size": "^4.0.0",
-    "sharp": "^0.23.2",
+    "sharp": "^0.25.2",
     "slash": "^2.0.0",
     "sockjs": "^0.3.19",
     "sockjs-client": "^1.3.0",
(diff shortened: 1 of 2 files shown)


hjvedvik:
Thanks, @marcvangend. Your PR looks good to me :) The only thing missing is updating the `yarn.lock` file. To do so, you can run `yarn` in the root folder. Just make sure you don't have any test projects in the `/projects` folder, or else their dependencies will also end up in the lock file.

I tried building gridsome.org with Sharp 0.25, and it seemed to work fine.

marcvangend:
Thanks @hjvedvik. I generated a new yarn.lock. It pulled in some other updates as well, but that was to be expected I guess.

hjvedvik:
I think its okay, @marcvangend :)

PR #1148, opened 2020-05-05
docs: update JAMstack -> Jamstack

eric-hc:
Updating to new Jamstack brand naming (jamstack.org has updated)

We should also update the repo description to `Jamstack` as well


hjvedvik:
Thanks, @eric-hc :)

Issue #1032, opened 2020-03-11
Hash did not match


ilicmarko:
## Description
There is an unpredictable behavior that we are getting from Gridsome. We are hosted on Netlify with a custom domain.

Sometimes, I still can't figure out the pattern, the whole website will break with an error:

```
app.b60ceaa8.js:formatted:22909 Error: Hash did not match.
    at s (app.b60ceaa8.js:formatted:5471)
    at app.b60ceaa8.js:formatted:5463
```
This is part of the code:

```
e.a = function(t) {
        var e = arguments.length > 1 && void 0 !== arguments[1] ? arguments[1] : {}
          , r = e.shouldPrefetch
          , l = void 0 !== r && r
          , h = (e.force,
        document.querySelector('meta[name="gridsome:hash"]').getAttribute("content"));
        return new Promise((function(e, r) {
            var p = "*" === t.name ? "/404" : t.path
              , f = t.meta.dataPath || Object(i.f)(p) + "/index.json"
              , u = Object(i.f)("/assets/data/") + f;
            return l && !o[f] ? (a[f] || (a[f] = Object(n.a)(u)),
            a[f].then((function() {
                return e()
            }
            )).catch((function() {
                return e()
            }
            ))) : (o[f] || (o[f] = c(u)),
            o[f].then((function(t) {
                t.hash !== h ? r(s("Hash did not match.", "INVALID_HASH")) : e(t)
            }
            )).catch(r))
        }
        ))
    }
```

The really strange part is there are not deploys in between. Basically, for some users, it works for some it doesn't. What is even stranger that the working and broken hash are the same.

```
Broken:
6f6ec468f0dbaf1648990284541132b62975331d

Working:
6f6ec468f0dbaf1648990284541132b62975331d
```

The cache inside JSON is also the same:

```
  "hash": "6f6ec468f0dbaf1648990284541132b62975331d",
```

### Steps to reproduce

I can describe clear steps because I can't reproduce it every time.

URL is: https://weareneopix.com

### Expected result

To work for all users.

### Actual result

The website breaks. Javascript stops working.

### Environment

```
  System:
    OS: Linux 5.3 Ubuntu 19.10 (Eoan Ermine)
    CPU: (8) x64 Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
  Binaries:
    Node: 10.17.0 - ~/.nvm/versions/node/v10.17.0/bin/node
    Yarn: 1.16.0 - /usr/bin/yarn
[...]

PR #1150, opened 2020-05-06
fix(app): add extra debugging output on hash mismatch failure

mtlynch:
Several clients are reporting non-deterministic app crashes caused by a mismatch in hashes (#1032). This adds additional context to the error to make it easier to diagnose the root cause.


hjvedvik:
Thanks, @mtlynch :)

PR #1173, opened 2020-05-17
Add Native Image Lazy Loading

RehanSaeed:
Adds native browser lazy loading.

I have to reference some images placed in the static folder, so I can add paths to them in meta tags using vue-meta (Would love to know if there was a better way around this). Lazy loading doesn't seem to work for these images. Using native browser lazy loading should work.

### DIFF
--- a/gridsome/app/components/Image.js
+++ b/gridsome/app/components/Image.js
@@ -28,6 +28,7 @@ export default {
 
     switch (typeof props.src) {
       case 'string':
+        attrs.isLazy = !isImmediate
         attrs.src = props.src
 
         break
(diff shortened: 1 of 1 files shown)


hjvedvik:
Thanks for the PR, @RehanSaeed. Images in the static folder are copied to the dist folder, so you should be able to reference an image at `/static/images/image.jpg` with a path like `/images/image.jpg`. And it's better to use a normal `img` element with `loading="lazy"` for images in the static folder.

But I think it would be great to support native lazy loading for `g-image`. But we should probably check if the browser supports it and fallback to `IntersectionObserver` if not. And the `immediate` attribute should be deprecated in favor of `loading="eager"`.

PR #1175, opened 2020-05-17
Added the option to set the max number of Posts

scottyzen:
```javascript
module.exports = {
  plugins: [
    {
      use: '@gridsome/source-wordpress',
      options: {
        baseUrl: 'WEBSITE_URL', // required
        maxNumberOfPosts: 32 // optional
      }
    }
  ]
}
```

### DIFF
--- a/packages/source-wordpress/index.js
+++ b/packages/source-wordpress/index.js
@@ -130,10 +130,10 @@ class WordPressSource {
       const restBase = this.restBases.posts[type]
       const typeName = this.createTypeName(type)
       const posts = getCollection(typeName)
-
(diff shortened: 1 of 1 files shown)


noxify:
@scottyzen thanks for the contributation.

With your changes, you will fetch all records from the API and then you're stopping after X elements.

Is it not better to reduce the API result via `?per_page`?
> WP API Doc: https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/

scottyzen:
Hey @noxify. I tried to use that first be couldn't get it working the way I wanted. Using per_page still returns all the posts. I could limit use per_page and limit the number of total pages. That might be a better solution

PR #1176, opened 2020-05-18
fix (Plugin API Hooks) - Promises

thetre97:
Update to use tapPromise instead of tapAsync.

I could only get the `onBootstrap` hook to work when I made this change, and I found it more helpful to use the promise style hooks. However, I can change the hook code to properly pass the async callback function  if you think it should be kept as `tapAsync`.


hjvedvik:
Thanks, @thetre97. I agree. It should be a `tapPromise` hook :)

PR #1193, opened 2020-05-26
tests: add new helper and increase coverage

jamesgeorge007:
- Added new helper method to be consumed across the entire test suite.
- Test case to ensure `create` command shows up an appropriate warning if a directory with the same name (which isn't empty) already exists in path.


hjvedvik:
Thanks, @jamesgeorge007

PR #1237, opened 2020-06-18
replace deprecated slug by title

sjousse:


hjvedvik:
Thanks for the PR, @sjousse. But the `:slug` isn't actually deprecated :) Gridsome used to auto-generate slugs based on the `node.title` field before v0.7. And in order to not break existing behavior immediately, we chose to show a deprecation warning and fallback to `:title` when no `node.slug` field existed. The example route in the readme assumes that the nodes will have slugs already.

PR #1252, opened 2020-06-29
Add missing ref exists functions

orblazer:
Hello,

This add missing loki operator `$refExists` and `$refListExists`.
That fix `Error: fun is not function` when we have an request like :
```gql
query {
  allProductCategory(filter: { parent: { id: { exists: false } } }) {
    totalCount
    edges {
      node {
        name
        parent {
          id
        }
      }
    }
  }
}
```

Thanks


hjvedvik:
Thanks, @orblazer :)

Issue #1048, opened 2020-03-17
line numbering config for vue-remark?


jhfoo:
## Summary
Installed vue-remark (0.1.10) and prismjs (0.3.0). Both work as expected. Looking to enable line numbering like gridsome-plugin-remark-prismjs-all. Tried adding the following (not working as expected):

// main.js
import 'prismjs/plugins/line-numbers/prism-line-numbers.css'

// .md file
~~~ sh{numberLines: true}
test line 1
test line 2
~~~

Brief explanation of the feature.

### Basic example
Include a basic example or links here.
Line numbering per prismjs site: https://prismjs.com/plugins/line-numbers/

### Motivation
Why are we doing this? What use cases does it support? What is the expected outcome?
To enable this and possibly other prismjs plugins.

  System:
    OS: freebsd
    CPU: (1) x64 Intel(R) Xeon(R) CPU           X5560  @ 2.80GHz
  Binaries:
    Node: 13.3.0 - /usr/local/bin/node
    npm: 6.12.1 - /usr/local/bin/npm
  npmPackages:
    @gridsome/remark-prismjs: ^0.3.0 => 0.3.0
    @gridsome/source-filesystem: ^0.6.2 => 0.6.2
    @gridsome/transformer-remark: ^0.5.0 => 0.5.0
    @gridsome/vue-remark: ^0.1.10 => 0.1.10
    gridsome: ^0.7.0 => 0.7.13

eman1986:
I'm interested in this myself

Issue #1035, opened 2020-03-11
@gridsome/remark-prismjs Should Support Line Numbers and Highlighting


RehanSaeed:
## Summary

PrismJS has plugins to add the ability to [show line numbers](https://prismjs.com/plugins/line-numbers/):

![image](https://user-images.githubusercontent.com/2270364/76443913-0f99e700-63bb-11ea-8f37-57b497951e24.png)

It can also [highlight lines of code](https://prismjs.com/plugins/line-highlight/):

![image](https://user-images.githubusercontent.com/2270364/76443972-293b2e80-63bb-11ea-84b9-ec2abe03457d.png)

### Motivation

Vuepress supports line numbers globally in it's `config.js` file and you can highlught lines by specifying ```js{4}.

jhfoo:
The code has not been updated since Oct 24, 2019. Not sure if this is priority now. gridsome-plugin-remark-prismjs-all on the other hand is very current (less than a month since last commit).

RehanSaeed:
Switched to `gridsome-plugin-remark-prismjs-all`. All is now well.

jhfoo:
Me too, except for that vue-remarks doesn't seem to turn on line numbers when married with gridsome-plugin-remark-prismjs-all..

djmtype:
Why do the docs for [remark-prism](https://github.com/gridsome/gridsome/tree/master/packages/remark-prismjs) clearly state line numbers can be displayed by including `showLineNumbers: true` to the options and importing the style path to main.js, but they don't render?

PR #1254, opened 2020-07-01
feature: add option to toggle line numbers alongside code

Mini-ghost:
I add option `showLineNumbers: true` to display line numbers for `@gridsome/remark-prismjs`, and also tried to update the doc

this feature be mentioned at #1035  and #1048 

## Example:

set in `gridsome.config.js` 

```js
{
  remark: {
    plugins: [
      [
        require('./script/remark-prismjs'),
        {
          // display line numbers alongside code
          showLineNumbers: true
        }
      ]
    ]
  }
}
```
or if wish to only show line numbers on certain code blocks, leave false and use the `{ numberLines: true | number }` syntax below

````markdown
```js { numberLines: 21 }
interface IntersectionObserver {
  readonly root: Element | null;
  readonly rootMargin: string;
  readonly thresholds: ReadonlyArray<number>;
  disconnect(): void;
  observe(target: Element): void;
  takeRecords(): IntersectionObserverEntry[];
  unobserve(target: Element): void;
}
```
````

that will output this

image

### DIFF
--- a/packages/remark-prismjs/README.md
+++ b/packages/remark-prismjs/README.md
@@ -74,6 +74,16 @@ If you'd like to disable highlighting of inline code blocks, pass `transformInli
 - Type: `Boolean`
 - Default: `false`
 
-If you'd like to add line numbers alongside the code. pass `showLineNumbers: true` in the plugin options
+If you'd like to add line numbers alongside the code. You can pass `showLineNumbers: true` in the plugin options and import a Prism CSS about line numbers to `main.js` file. like this:
 
-If you wish to only show line numbers on certain code blocks, you can leave `false` and use the `{ numberLines: true }` syntax below
+```js
+import 'prismjs/themes/prism.css'
+// Prism default CSS about line numbers
+import 'prismjs/plugins/line-numbers/prism-line-numbers.css'
+
+export default function (Vue) {
+  // ...
+}
+```
+
+If you wish to only show line numbers on certain code blocks, you can leave `false` and use the `{ lineNumbers: true }` syntax below
--- a/packages/remark-prismjs/README.md
+++ b/packages/remark-prismjs/README.md
@@ -60,23 +60,20 @@ module.exports = {
 }
 ```
 
+## Options
+
+#### transformInlineCode
+
+- Type: `Boolean`
+- Default: `false`
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by hjvedvik** [comment]
Thanks, @Mini-ghost! That's a feature that we've wanted for a while :) It seems to work well, but I think it maybe needs a minor change to work well with the default CSS. The `.line-numbers-rows` element should probably be a direct child of the `code` element instead.

And should we mention how to include the default CSS for line numbers in the readme?

```js
import 'prismjs/plugins/line-numbers/prism-line-numbers.css'
```

Mini-ghost:
@hjvedvik Oops! I accidentally missed to mention how to include the default CSS for line numbers in the readme. I will add this paragraph.

About `.line-numbers-rows` element, now it is instead `pre` element (behind `code` element), like this:

```html
<pre class="language-typescript line-numbers">
  <code class="language-typescript">
    <!-- code contain  -->
  </code>
  <span class="line-numbers-rows" aria-hidden="true">
    <!-- span element of the line number -->
  </span>
</pre>
```

And your mean is this?

```html
<pre class="language-typescript line-numbers">
  <code class="language-typescript">
    <!-- code contain  -->
    <span class="line-numbers-rows" aria-hidden="true">
      <!-- span element of the line number -->
    </span>
  </code>
</pre>
```

hjvedvik:
> And your mean is this?

Yes, that's right :) At least that's how the markup is in the [Line Numbers](https://prismjs.com/plugins/line-numbers/) plugin examples. And I could only get the default CSS to work with that markup.

And just one thought. Should we name the inline option `lineNumbers` instead? Just so it's more consistent with the global `showLineNumbers` option and the PrismJS plugin name.

```markdown
```js { lineNumbers: true }
interface IntersectionObserver {
  readonly root: Element | null;
```

Mini-ghost:
> And just one thought. Should we name the inline option lineNumbers instead? Just so it's more consistent with the global showLineNumbers option and the PrismJS plugin name.

You are right. After I tried to use this option a many times, inconsistent naming is a little confusing for me , so I also think consistent naming is better.

I will adjust the structure of HTML elements, consistent option naming, and update the readme.

hjvedvik:
The PR looks good now :) Thanks!

PR #1257, opened 2020-07-03
Fix `g-image` not adding `height` attribute

hacknug:
This PR fixes an issue where the browser wouldn't be able to save space for images even when this info was available 😢 

> Use both width and height to set the intrinsic size of the image, allowing it to take up space before it loads, to mitigate content layout shifts. 
> [``: The Image Embed element - HTML: Hypertext Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)

### DIFF
--- a/gridsome/app/components/Image.js
+++ b/gridsome/app/components/Image.js
@@ -40,6 +40,7 @@ export default {
 
         attrs.src = isLazy ? dataUri : src
         attrs.width = size.width
+        attrs.height = size.height
 
         if (isLazy) attrs['data-src'] = src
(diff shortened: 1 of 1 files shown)


djmtype:
Every Gridsome project utilizing g-image has content shift issues because this one simple line is missing. Please merge.

NickSto:
Just adding a kludge I found on the discord: @soulfulwater made a [patch](https://gist.github.com/soulfulwater/7a54244187cb0023f6c0117f1c39964c) you can apply using [patch-package](https://www.npmjs.com/package/patch-package) to fix this, until Gridsome puts out another release.

PR #1258, opened 2020-07-04
feat(source-wordpress) WordPress source updates

thetre97:
I am in the process of updating this plugin, to remove deprecated functionality (like `route` in `addCollection`) and to try and fix some current issues. It also has the functionality to download images, for use with `g-image`.


thetre97:
I have made all the updates I planned, and tried to fix any outstanding issues etc. It will no doubt need testing though.

As for WooCommerce support, I am working on that in a test repo, and will add when ready - it will be quite a major change though.

PR #1264, opened 2020-07-08
fix(vue-remark): add support VueComponent function

sebtiz13:
Hello,

When create component with vue-class-component or Vue.extend
import return an VueComponent function not an Object.

So, i have added an branch in if to support this case

I test and don't change original behavior and work fine with the two form.
Also work fine with components in typescript (with vue-class-component in my case)


hjvedvik:
Thanks, @sebtiz13 :)

PR #1282, opened 2020-07-22
fix: prefer npm if the starter template has package-lock.json

### LINKED ISSUE #1273: gridsome create doesn't use npm with package-lock.json
## Description

I'm using `gridsome create` with a starter that used npm and has a `package-lock.json`, but it looks like gridsome installs using yarn because I end up with a `yarn.lock` file.

### Steps to reproduce

1. `gridsome create gridsome-npm-example git@github.com:srchulo/gridsome-npm-example.git`
2. Check for `yarn.lock`

Here's the repo:

https://github.com/srchulo/gridsome-npm-example

### Expected result

`yarn.lock` should not exists

### Actual result

`yarn.lock` exists, which I assume means yarn was used to install the dependencies.

### Environment

```
  System:
    OS: Linux 4.19 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-6600K CPU @ 3.50GHz
  Binaries:
    Node: 14.5.0 - /usr/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.6 - /usr/bin/npm
  npmPackages:
    gridsome: ^0.7.19 => 0.7.19
  npmGlobalPackages:
    @gridsome/cli: 0.3.3
```

jamesgeorge007:
fixes #1273

### DIFF
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -60,7 +60,7 @@ module.exports = async (name, starter = 'default') => {
       title: `Install dependencies`,
       task: (_, task) => {
         let command = 'npm'
-        if (!fs.existsSync(`${dir}/package-lock.json`)) {
+        if (!fs.existsSync(path.join(dir, 'package-lock.json'))) {
           command = useYarn ? 'yarn' : 'npm'
         }
         const stdio = ['ignore', 'pipe', 'ignore']
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -59,7 +59,10 @@ module.exports = async (name, starter = 'default') => {
     {
       title: `Install dependencies`,
       task: (_, task) => {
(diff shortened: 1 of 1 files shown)


packages/cli/lib/commands/create.js
```diff
@@ -59,7 +59,10 @@ module.exports = async (name, starter = 'default') => {
     {
       title: `Install dependencies`,
       task: (_, task) => {
-        const command = useYarn ? 'yarn' : 'npm'
+        let command = 'npm'
+        if (!fs.existsSync(`${dir}/package-lock.json`)) {
```
hjvedvik, line 63:
```suggestion
        if (!fs.existsSync(path.join(dir, 'package-lock.json'))) {
```
Using template literals to merge paths caused some issues on Windows for me a while ago. So I think we should use the `path` module to get the correct path. But I don't have Windows available atm, so I can't test it...

Issue #1195, opened 2020-05-28
Question regarding images in markdown


johannesholmberg:
## Description

Adding an image in a markdown file adds the image nicely to the document, but it wraps it inside a `<p>` tag. I was wondering if it’s possible to remove that wrapping? I would like to have the `<img>` plain the markup so I can target it with css.

noxify:
Hi,

i had the same issue and solved it by disabling the built-in image processing for images.

I copied the original code from the image processor and created my own remark plugin.
This is just an dirty hack, but if you like it dirty, you can checkout my repo:
* https://github.com/noxify/gridsome-starter-liebling/tree/master/packages/gridsome-plugin-remark-figure
* https://github.com/noxify/gridsome-starter-liebling/blob/master/gridsome.config.js#L73-L76

But we have already an open issue for this topic: https://github.com/gridsome/gridsome/issues/700

The workaround from @sebtiz13 seems to be a good alternative: https://github.com/gridsome/gridsome/issues/700#issuecomment-632888229

sebtiz13:
Hi

@noxify i have look your custom plugin and i wonder if property like `wrapper` in core plugin can be an solution for your case and unwrapping for moment i have just create a lazy solution (https://github.com/sebtiz13/gridsome/commit/eab58a0dd324d055e277f494413a7753278c7426) with `unwrapImages` property

I think like this:
```JS
{
  wrapperImages: 'figure', // replace p by figure
  wrapperImages: false, // just remove wrapper
}
```

But i have two questions maybe to debate to potentially improve the solution:
* use `classNames` it's used in wrapper instead of img (figure for example)
* `wrapperImages` can accept a function to manage a specific case like yours

Issue #700, opened 2019-09-23
Add remark-unwrap-images to @gridsome/transformer-remark default plugins


missmatsuko:
## Summary
`remark-unwrap-images` remark plugin removes paragraph tags that wrap images.

### Basic example
[GitHub for remark-unwrap-image](https://github.com/remarkjs/remark-unwrap-images)

### Motivation
It seems like a good default plugin to add since paragraphs shouldn't be wrapping images.

midlantica:
Is there no way to break this curse? :P

missmatsuko:
Right? I tried adding it myself but couldn't get it working. I think there's some conflict between when images get unwrapped and when they get replaced with g-image.

kevinzweerink:
Also interested in support for this—didn't work for me either when I tried to install it myself. Currently using Vue lifecycle hooks to add classes to images within paragraphs, but it doesn't feel like a good solution.

jamesjlyons:
I'm also trying to get this to work. I thought this might have to do with #707 and the new vue-remark, but it seems like a separate issue for me since clearing my cache and trying multiple configurations in my gridsome.js hasn't helped.

AchrafTani:
@missmatsuko Isn't there anything new about this? I'm also having the same problem ? 
Thanks.

sebtiz13:
hello remark-unwrap-image can't unwrap the image transformed by gridsome because it's convert in  `html` block.

https://github.com/gridsome/gridsome/blob/master/packages/transformer-remark/lib/plugins/image.js#L42-L45

But this plugin search block of type `image` or `imageReference`.
I think it's convert in `html` block because have "prerendered" img and noscript element.

A good solution can be add an option like "unwrapImage".
I have try this solution with simple implementation of plugin.
https://github.com/sebtiz13/gridsome/commit/f6f25e9fd2cd6d0a50dad0cd0439b6d41727c318

In waiting of real solution, i have create an hacky plugin to unwrap the outputted html image without modify package.

https://gist.github.com/sebtiz13/22c2765699773fcc45e017d521704c71

alexwidua:
In case someone runs into the same issue, I did a quick Gridsome implementation of [gatsby-remark-unwrap-images](https://github.com/cedricdelpoux/gatsby-remark-unwrap-images), you can find the repo here: [gridsome-remark-unwrap-images](https://github.com/alexwidua/gridsome-remark-unwrap-images).

[...]

PR #1290, opened 2020-07-25
fix(remark): add option to control wrapper of images (#700, #1195)

sebtiz13:
add an option `wrapperImages` for [@gridsome/transformer-remark](https://gridsome.org/plugins/@gridsome/transformer-remark) to control the wrapper of images.

`wrapperImages` can be:

- `false`: simply remove wrapper
- `string`: name of tag to wrap images (ex: figure)
- `function`: the function permit all custom modify

the `function` receive:

- **node**: node tree of paragraph
- **index**: index of node in tree
- **parent**: the parent of node

# example
``` javascript
{
  use: '@gridsome/source-filesystem',
    options: {
    path: 'content/**/*.md',
      typeName: 'BlogPost',
        remark: {
      wrapperImages: (node) => {
        node.type = 'figure'
        node.data = {
          hNames: 'figure',
          hProperties: {
            className: 'figure-image'
          }
        }
      }
    }
  }
}
```

This configuration render something like this in html
```html
<figure class="figure-image">
  <img src="myImage.png">
</figure>
```

### DIFF
--- a/packages/transformer-remark/lib/plugins/image.js
+++ b/packages/transformer-remark/lib/plugins/image.js
@@ -9,7 +9,31 @@ module.exports = function attacher (options = {}) {
 
     const images = []
 
-    visit(tree, 'image', node => images.push(node))
+    if (typeof options.wrapper !== 'undefined') {
+      visit(tree, 'paragraph', (node, index, parent) => {
+        let hasImage = false
+        node.children.forEach(children => {
+          if (children.type === 'image') {
+            hasImage = true
+            images.push(children)
+          }
+        })
+        if (hasImage) {
+          if (options.wrapper === false) {
+            parent.children.splice(index, 1, ...node.children)
+          } else if (typeof options.wrapper === 'string') {
+            node.type = options.wrapper
+            node.data = {
+              hName: options.wrapper
+            }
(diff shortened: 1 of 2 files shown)


alexwidua:
I‘m curios, does it preserve the noscript tag of g-images? And if yes, does the noscript also gets wrapped?

sebtiz13:
@alexwidua yes it preserve the noscipt tag i forgot in my example but it work :wink: 
Yes noscript is still wrapped with g-images like default.

How it's work ?

https://github.com/gridsome/gridsome/blob/679c57175eaed4dc8284c62a109c299a3bf24b18/packages/transformer-remark/lib/plugins/image.js#L18
In visit function the images objects, it's added in array

https://github.com/gridsome/gridsome/blob/679c57175eaed4dc8284c62a109c299a3bf24b18/packages/transformer-remark/lib/plugins/image.js#L68
After all images it's compiled and replaced by reference in html object with g-image + nosript in value

So since the image is saved before remove or alter wrapper we have no risk to alter the render of g-image.
Only flaw is if you delete an image in children, it is still compiled but not rendered.
This is the only risk, but why delete the image in this process :no_good: ?

PR #1308, opened 2020-08-13
Source-strapi media download

upalatucci:
Hi guys,

In so excited because this is my first contribution to an open-source project.

In my company, we used Gridsome combined with Strapi and we noticed that the images stored in Strapi are not downloaded from the remote server using the source-strapi package.
So I thought to create a custom code to do the media download in the api.loadSource.

I'm absolutely open to criticism as I want to learn as much as I can from the community.
What do you think about this feature? 

I had to adapt my script to be used into the package, but I tested it and works fine with images.

For this feature, I add two arguments in the options that can be added also in the doc if you agree with the introduction of this feature. 

mediaDownloadFolder: the absolute path of the destination directory 
(Can be in some way relative path? I did not know how can be implemented with relative path)

mediaDownloadParameters (can have a better name?): the key names of the media in the root that has to be downloaded. 

For now, only images in the object root are detected but can be easily implemented the code for detect parameters along the object.
For example can be implemented a sintax like "object1.object2.imageToBeDownloaded".

Sorry for my poor explanation. Hope I was clear. English is not my native language.

### DIFF
--- a/packages/source-strapi/helpers/query.js
+++ b/packages/source-strapi/helpers/query.js
@@ -1,8 +1,8 @@
 const axios = require('axios')
 const pluralize = require('pluralize')
 const execSync = require('child_process').execSync
-const path = require("path")
-const fs = require("fs")
+const path = require('path')
+const fs = require('fs')
 
 function downloadMedia (image, apiUrl) {
   const { url, sha256, ext } = image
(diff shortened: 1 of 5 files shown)


hjvedvik:
Thanks for the PR, @upalatucci! I can take a look at it one of the next days :)

packages/source-strapi/helpers/query.js
```diff
@@ -23,6 +66,15 @@ module.exports = async ({ apiURL, resourceName, jwtToken, queryLimit, isSingleTy
   // Make API request.
   return axios(apiEndpoint, fetchRequestConfig)
     .then(res => res.data)
+    .then(async data => {
+      if (mediaDownloadFolder) {
+        await execSync(`mkdir -p ${mediaDownloadFolder}`)
```
hjvedvik, line 71: Would it be better to use the [`ensureDir()`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md) method from the `fs-extra` module here instead?

upalatucci: Yeah great!

upalatucci:
Sorry @hjvedvik  for all these days for a simple commit.

Issue #551, opened 2019-07-14
Use Gridsome with pnpm


Js-Brecht:
I just wanted to drop this here in case anybody else is trying to use Gridsome, while using pnpm as their package manager.
* There are some issues with Webpack resolving modules, but that seems to be pretty common when using pnpm, because of the method that webpack uses.
* Gridsome also makes a few assumptions for you about some Vue packages
   * I'm assuming this is because they want users to be locked in to a specific version of the Vue library, to ensure compatibility with the Gridsome framework.
   * As a result, they are not among your project's dependencies in a gridsome scaffolded project.

Btw, this is after scaffolding a new project with `@gridsome/cli`:
```sh
gridsome create <project-name>
```

First thing, we have to tell Webpack how to find its modules.  In your project root folder, you need to include this in `gridsome.server.js`:
```js
const fs = require('fs');
const path = require('path');

let gridsomeModules = fs.realpathSync("node_modules/gridsome");
gridsomeModules = path.resolve(gridsomeModules, "..");

module.exports = function (api) {
  ...
  api.chainWebpack((config) => {
    config.resolveLoader.modules.prepend(gridsomeModules);
    return config;
  });
  ...
}
```

Then there's a few dependencies that need to be loaded.  
* Gridsome uses babel, and babel wants `core-js` somewhere, but it's not available in Gridsome's modules.
   * Needs `"core-js": "^2.6.9"`
      * Note: I had to use version `^2.6.9` because version 3+ doesn't have the es6 modules where babel is expecting them.

Make a pnpmfile.js, and put this hook into it:
```js
// pnpmfile.js
function readPackage(pkg, context) {
  switch(pkg.name) {
    case "gridsome":
      pkg.dependencies = Object.assign({}, pkg.dependencies, {
        "core-js": "^2.6.9",
      });
      break;
  }
  return pkg;
}

module.exports = {
  hooks: {
    readPackage,
  },
};
```

* Because pnpm isn't using a flat node_modules directory like npm/yarn, the libraries you need to develop using Vue are not available to you, so they need to be installed.  Your project needs the following packages:
   * Because of their nature, I put these in `dependencies`:
      * `"vue": "^2.6.10"`
      *  `"vue-meta": "^2.0.5"`
      * `"vue-router": "^3.0.7"`
      * `"vue-template-compiler": "^2.6.10"`

These particular dependencies appear to be needed due to Webpack's quirks.
   * Because they are for building/testing, I put them in `devDependencies`:
[...]

PR #1313, opened 2020-08-16
feat: support Yarn PnP

merceyz:
**What's the problem this PR addresses?**

- Gridsome doesn't work with [Yarn PnP](https://yarnpkg.com/features/pnp).
- Gridsome is passing loaders to webpack using their name instead of absolute path
- Gridsome has some undeclared dependencies

Fixes https://github.com/gridsome/gridsome/issues/551 (cc @Js-Brecht)
Closes https://github.com/yarnpkg/berry/issues/1266 (cc @borisdiakur)

**How did you fix it?**

- Use `require.resolve` for webpack loaders (https://github.com/yarnpkg/berry/tree/eff8ffd607242d80adbeaa05ebf5453a68dd299b/packages/yarnpkg-doctor#no-unqualified-webpack-config)
- Add undeclared dependencies (https://github.com/yarnpkg/berry/tree/eff8ffd607242d80adbeaa05ebf5453a68dd299b/packages/yarnpkg-doctor#no-unlisted-dependencies)
- Resolve user defined plugins relative to the context (gridsome.config.js)
- Add https://github.com/arcanis/pnp-webpack-plugin (the only PnP specific thing in this PR)
- Added an E2E test to make sure all dependencies are declared and required correctly

### DIFF
--- a/gridsome/lib/webpack/plugins/corejsBabelPlugin.js
+++ b/gridsome/lib/webpack/plugins/corejsBabelPlugin.js
@@ -1,0 +1,13 @@
+const corejsPath = require('path').dirname(require.resolve('core-js/package.json')).replace(/\\/g, '/')
+
+module.exports = () => ({
+  visitor: {
+    ImportDeclaration(path) {
+      if (path.node.specifiers.length !== 0 || path.node.source.value.indexOf('core-js/') !== 0) {
+        return
+      }
+
+      path.node.source.value = corejsPath + path.node.source.value.substr('core-js'.length)
+    },
+  },
+})
--- a/gridsome/lib/webpack/plugins/corejsBabelPlugin.js
+++ b/gridsome/lib/webpack/plugins/corejsBabelPlugin.js
@@ -8,6 +8,6 @@ module.exports = () => ({
       }
 
       path.node.source.value = corejsPath + path.node.source.value.substr('core-js'.length)
-    },
-  },
+    }
+  }
 })
--- a/gridsome/package.json
+++ b/gridsome/package.json
@@ -103,6 +103,22 @@
     "webpack-merge": "^4.2.1",
     "yaml-loader": "^0.5.0"
   },
+  "peerDependencies": {
+    "less-loader": "*",
+    "sass-loader": "*",
+    "stylus-loader": "*"
+  },
+  "peerDependenciesMeta": {
+    "sass-loader": {
+      "optional": true
+    },
+    "less-loader": {
+      "optional": true
+    },
+    "stylus-loader": {
+      "optional": true
+    }
+  },
   "engines": {
     "node": ">=8.3"
   },
--- a/gridsome/package.json
+++ b/gridsome/package.json
@@ -83,6 +83,7 @@
     "portfinder": "^1.0.20",
     "postcss-loader": "^3.0.0",
     "probe-image-size": "^4.0.0",
+    "resolve-from": "^5.0.0",
     "sharp": "^0.25.2",
     "slash": "^2.0.0",
     "sockjs": "^0.3.19",
--- a/gridsome/package.json
+++ b/gridsome/package.json
@@ -79,6 +79,7 @@
     "p-map": "^2.0.0",
     "path-to-regexp": "^2.2.1",
     "physical-cpu-count": "^2.0.0",
+    "pnp-webpack-plugin": "^1.6.4",
     "portfinder": "^1.0.20",
     "postcss-loader": "^3.0.0",
     "probe-image-size": "^4.0.0",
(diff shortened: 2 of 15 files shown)


gridsome/lib/webpack/createServerConfig.js
```diff
@@ -11,7 +11,6 @@ module.exports = async app => {
   config.entry('app').add(resolve('../../app/entry.server.js'))
 
   config.target('node')
-  config.externals([/^(vue|vue-router|vue-meta)$/])
```
merceyz, line 14: Making these external causes `vue-server-renderer` to require them, which is wrong as it doesn't declare them as dependencies.
I couldn't figure out how to make `vue-server-renderer` require them on behalf of `gridsome` so it removed for now

gridsome/package.json
```diff
@@ -101,6 +104,22 @@
     "webpack-merge": "^4.2.1",
     "yaml-loader": "^0.5.0"
   },
+  "peerDependencies": {
+    "less-loader": "*",
+    "sass-loader": "*",
+    "stylus-loader": "*"
```
merceyz, line 110: The assumption here is that the user needs to provide these

gridsome/lib/webpack/createServerConfig.js
hjvedvik: I don't think it will be a problem to include them in the server bundle, for now. I've been working on implementing Vue 3 lately and made a simple bundle renderer where the [resolver](https://github.com/vuejs/vue/blob/dev/src/server/bundle-renderer/create-bundle-runner.js#L53-L65) is run within `gridsome`. That should make it possible to have them as externals again, I think.

And thanks for the PR. I'll take a closer look one of the next days!

hjvedvik: It seems like `createBundleRenderer` from `vue-server-renderer` can take a `basedir` option. Which will make `vue-server-renderer` resolve the exernal from that dir instead. You can try to add `basedir: __dirname` [here](https://github.com/gridsome/gridsome/blob/master/gridsome/lib/server/createRenderFn.js#L19-L24) and see if it helps.

merceyz: Ah there it is, that works. :+1:

packages/source-graphql/package.json
```diff
@@ -17,7 +17,8 @@
     "node-fetch": "^2.6.0"
   },
   "peerDependencies": {
-    "gridsome": "0.x"
+    "gridsome": "0.x",
+    "graphql": "^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0"
```
hjvedvik, line 21: Can this be a direct dependency instead? `apollo-link-http` and `graphql-tools` have it as a peer dependency, and this package is requiring GraphQL types from `gridsome/graphql`. So it would be nice if the users don't need to install `graphql` in their projects.

merceyz: That would be possible, but then there is no guarantee that there will be one graphql version/instance. Though there isn't with this either as the user can install something different than what `gridsome` uses.

merceyz: What would you like to do here?

hjvedvik: I think we can add it as a direct dependency here. And maybe be more specific about the `gridsome` version listed as peer dependency to ensure this package uses the same `graphql` version as `gridsome`. Do you think that will work?

merceyz: It has a posibility to work, but it's not guaranteed to be the same version

merceyz: I've made it a direct dependency

hjvedvik:
I noticed the e2e test passed successfully. Great job @merceyz :) Do you think this PR is ready to be merged?

merceyz:
> Do you think this PR is ready to be merged?

Yup, it's ready

jagu-sayan:
Any news on this great feature ?

hjvedvik:
@merceyz Do you have some time to resolve the conflict and update this PR? I would like to rebase-merge these commits to preserve the individual changelogs for each package. Sorry for taking this long to get back to you...

gridsome/lib/webpack/plugins/corejsBabelPlugin.js
```diff
@@ -0,0 +1,13 @@
...
+  visitor: {
+    ImportDeclaration(path) {
+      if (path.node.specifiers.length !== 0 || path.node.source.value.indexOf('core-js/') !== 0) {
+        return
+      }
+
+      path.node.source.value = corejsPath + path.node.source.value.substr('core-js'.length)
+    }
+  }
+})
```
merceyz, line 13: This should be replaced with https://github.com/babel/babel-polyfills/tree/main/packages/babel-plugin-polyfill-corejs3 as doing this naive find and replace can break stuff really easily.

Unsure which version of `@babel/core` or `@babel/preset-env` is needed for it to work (cc @nicolo-ribaudo) so if you'd like to update the babel dependencies in a seperate commit that would probably be best

Issue #232, opened 2019-03-01
GraphQL Fragments


sanscheese:
## Summary

When wanting to display a list of items in different template can find queries are being repeated. [GraphQL fragments](https://graphql.org/learn/queries/#fragments) should help keep things DRY.

### Basic example
Not sure where the fragment would best be placed, or if it would be developers choice
``` GraphQL
fragment postFields on Post {
    title
    path
    date (format: "D. MMMM YYYY")
    timeToRead
    description
    tags {
      id
      title
      path
    }
}
```

Used in `<page-query>` of `src/template/Post.vue` 
``` GraphQL
query Post ($path: String!) {
  post: post (path: $path) {
    ...postFields
    coverImage (width: 1060, blur: 10)
  }
}
```

Then used in `<page-query>` of `src/template/Tag.vue` 
``` GraphQL
query Tag ($id: String!) {
  tag (id: $id) {
    title
    belongsTo {
      edges {
        node {
          ...on Post {
            ...postFields
            coverImage (width: 360, blur: 10)
          }
        }
      }
    }
  }
}
```

### Motivation

Keep thing DRY and easier to maintain if wanting to add/remove fields to content types in future.

hjvedvik:
Fragments would be great! We could include them automatically by loading them from a specific folder or file. Or let developers import them into each query. Something like this package: https://github.com/prisma/graphql-import, but I'm not sure if it works with fragments, so we would need to create a solution for it.

```graphql
# import DefaultPost from "~/fragments/post.graphql"

query Post ($path: String!) {
  post: post (path: $path) {
    ...DefaultPost
    coverImage (width: 1060, blur: 10)
  }
}
```

ahsath:
Definitely a good feature to have

davestewart:
Hello. Did anything happen with this?

sanscheese:
Still tbc. I hoping to dig into it a bit to see if I can work out a solution (but not much time to give, for next little while...)

hacknug:
@hjvedvik Apollo supports importing fragments from `.graphql` files: https://www.apollographql.com/docs/react/integrations/webpack/#fragments

This would be an amazing addition since right now we're forced to either:
- duplicate queries and fragments in pages and components, or
- use `$fetch` to get the data queried in those pages affecting the site's performance

adamJLev:
Would be great to have this.
What gatsby does looks good to me: https://www.gatsbyjs.org/docs/using-graphql-fragments/

[...]

PR #1329, opened 2020-09-04
Support GraphQL fragments

sebtiz13:
Like I say in #232, I take inspired by the implementation of Gatsby.

Actual features:

- He get the defined fragments in components and layouts files
- All fragments in this files are global and can be used in Pages and Templates vue files.
- Check unused fragments
- Check duplicate fragments names
- All errors are reported in browser with webpack

Just one error can't be reported in browser it's "unknown fragments ..." when delete fragment in a component. Because page request are updated, but pages and templates are not re compiling with webpack so he not pass by her validation.

A solution like purposed @hjvedvik (https://github.com/gridsome/gridsome/issues/232#issuecomment-468627300) could be also implemented, but I find this approach better to allow to have the fragments in components where he are used.
In my case i would like create an fragments to get my menu so fragments can be stored in my SFC component menu.
(Why i would use fragments instead of static-query ? because static-query don't support variables)

How to use:
An simple example with my case
```vue
// components/menu.vue
<template>
<a v-for="(menuLink, key) in $page.menuLink" :key="key" href="menuLink.link">{{ menuLink.name }}</a>
</template>
<page-query>
fragment menuLink on Query {
  menuLink(locale: $locale) {
    link
    name
  }
}
</page-query>
```

```vue
// pages/index.vue
<template>
<menu />
</template>
<page-query>
query Index($locale: String = 'en') { // Off course, $locale can be dynamic with the URL or other
  ...menuLink
}
</page-query>
```

I hope it can be useful and all feedback are welcome.


jmdmacapagal:
bumping on this

borisbrue:
bumping on this ;)

Issue #748, opened 2019-10-11
@gridsome/source-contentful locales not considered.


sliktrik:
In order to retrieve locale-based content with Contentful, the locale i.e. 'en' needs to be passed as part of the GraphQL.

Currently, it doesn't appear that locale has been considered as part of the '@gridsome/source-contentful' package. Only 'id' and 'path' are available.

bircun:
I'm new on Gridsome, and first I would like to thanks the Gridsome team and contributors for doing such a great job 👍

About this issue that @sliktrik mentioned, I have the same problem. I'm building a website in 2 locales using Contentful as data source, and only the "default" locale is imported and available in Gridsome GraphQL. For example in Gatsby, there is a "node_locale" field that comes, that we can use for filtering the nodes.

Do you plan to add the multi locales support soon? Thanks

sangeet:
Had similar issues in the past. Had to move the project to Gatsby. I'd like to know if anyone has a solution to this problem. Basically makes Contentful sites with multiple locales unusable with Gridsome.

smonist:
I have spent way too much time figuring this out. Please add support for locales or at least state that there is none.

helloiamlukas:
any updates on this?

helloiamlukas:
I created [a pull request](https://github.com/gridsome/gridsome/pull/1331) now that solves this problem. 

Further I'm working on a plugin that automatically creates localized pages (similar to [gridsome-plugin-i18n](https://github.com/daaru00/gridsome-plugin-i18n)) from the Contentful source. As soon as I'm done I will post it here. ☺️

helloiamlukas:
I ended up adding multi-language support for the Contentful source now, see (#1341). You can use it in combination with [gridsome-plugin-i18n](https://github.com/daaru00/gridsome-plugin-i18n), works perfectly together.

93lucasp:
any news?

helloiamlukas:
@93lucasp As already mentioned in my previous post, this problem is pretty much solved [here](https://github.com/gridsome/gridsome/pull/1341)

PR #1331, opened 2020-09-06
fix(source-contentful) add possibility to define search parameters in config

helloiamlukas:
This will solve #748 

Previously the Contentful source only returned the default locale of a field. As [mentioned in the Contentful docs](https://www.contentful.com/developers/docs/references/content-delivery-api/#/reference/localization), the search parameter `locale=*` has to be provided to return all locales of a field. 

With the changes in this pull request, you can now define search parameters in the plugin configuration. 

```
{
   options: {
      space: process.env.CTF_SPACE_ID,
      accessToken: process.env.CTF_ACCESS_TOKEN,
      host: 'cdn.contentful.com',
      environment: 'master',
      typeName: 'Contentful',
      parameters: {
         locale: '*'
      }
   }
}
```

### DIFF
--- a/packages/source-contentful/index.js
+++ b/packages/source-contentful/index.js
@@ -111,7 +111,7 @@ class ContentfulSource {
   }
 
   async fetch (method, limit = 1000, order = 'sys.createdAt', parameters = {}) {
-    const fetch = skip => this.client[method]({ skip, limit, order, ...parameters })
+    const fetch = skip => this.client[method]({ ...parameters, skip, limit, order })
     const { total, items } = await fetch(0)
     const pages = Math.ceil(total / limit)
 
--- a/packages/source-contentful/index.js
+++ b/packages/source-contentful/index.js
@@ -11,7 +11,8 @@ class ContentfulSource {
       host: 'cdn.contentful.com',
       typeName: 'Contentful',
       richText: {},
-      routes: {}
+      routes: {},
+      parameters: {}
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/source-contentful/index.js
```diff
@@ -108,8 +110,8 @@ class ContentfulSource {
     }
   }
 
-  async fetch (method, limit = 1000, order = 'sys.createdAt') {
-    const fetch = skip => this.client[method]({ skip, limit, order })
+  async fetch (method, limit = 1000, order = 'sys.createdAt', parameters = {}) {
+    const fetch = skip => this.client[method]({ skip, limit, order, ...parameters })
```
hjvedvik, line 114:
```suggestion
    const fetch = skip => this.client[method]({ ...parameters, skip, limit, order })
```
Just so `parameters` doesn't override the internal parameters.

packages/source-contentful/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "version": "0.5.3",
-  "name": "@gridsome/source-contentful",
+  "name": "@helloiamlukas/source-contentful",
```
hjvedvik, line 3:
```suggestion
  "name": "@gridsome/source-contentful",
```

helloiamlukas: sorry, this never belonged here 😔  reverted it now

hjvedvik:
Looks good now, @helloiamlukas :) Thanks for the PR!

NicholasRowe:
What is the latest with #1341 ?

Issue #715, opened 2019-09-29
Surface graphql function to plugins


cco3:
It'd be nice if the api gave access to the graphql function.

### Basic example

```
api.afterBuild(async ({ config, graphql }) => {
  const data = await generatePodcast(graphql)
  const outFile = path.join(config.outDir, 'podcast.xml')
  await fs.writeFile(outFile, data)
})
```

### Motivation

It seems like several of the api functions could benefit from access to the actual site data, including the podcast generating example above.

hjvedvik:
Yes, it would be useful to have access to the GraphQL schema after the build. I'll add this to the todo-list :)

platform-kit:
@hjvedvik Is there another thread for discussing this idea, or shall we do it here?

I would love to write a plugin that exposes a Graphql API, ideally via a serverless script so it could be easily (zero-config) deployed on Netlify. Just not sure where to start.

PR #1333, opened 2020-09-07
Make graphql function available in afterBuild

cco3:
Issue: #715


hjvedvik:
Thanks for the PR, @cco3. My only concern with this change is that the store has to be in memory until the `api.afterBuild()` hook has run. I think we are going to remove the store right after the JSON-files have been generated in one of the next releases. But maybe we should add a new hook that runs either before or right after the "Execute GraphQL" step? For example a `api.onExecuteQueries()` hook?

PR #1348, opened 2020-09-22
Add custom class prefix to remark-prismjs

BernardZhao:
Prismjs's default classes for syntax highlighting can come into conflict with your own classes. Adding this plugin from prismjs lets us prefix its classnames so that we can avoid this. 

Before:
Screenshot 2020-09-21 235321

After:
Screenshot 2020-09-21 235418

I've ran into this issue myself and you can see examples of this causing issues even in Gridsome starters like [this one](https://calebanthony.github.io/gridsome-bulma/say-hello-to-gridsome). Notice how the code block looks weird, with anomalies like the giant number 5? That's from the conflicting number class from Bulma and number in Prismjs. I wonder if there should even be a prefix by default, if this really is so common, but I have implemented this as an option.

### DIFF
--- a/packages/remark-prismjs/README.md
+++ b/packages/remark-prismjs/README.md
@@ -62,6 +62,14 @@ module.exports = {
 
 ## Options
 
+#### customClassPrefix
+
+- Type: `String`
+- Default: `''`
+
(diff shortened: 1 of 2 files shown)


hjvedvik:
Thanks, @BernardZhao. Seems like a useful option to me.

Issue #1269, opened 2020-07-12
<g-link> doesn't support v-html property


mensch:
## Description

I couldn't decide whether this is a bug report or a feature request, but he Gridsome documentation states that g-link is a wrapper for router-link so I'm leaning towards the former. I'm using g-link in combination with some records retrieved via GraphQL, so the link label comes from an external datasource. Wrapping the label in the g-link tag has the effect of escaping any special characters as per standard Vue behaviour. I'd like to leverage the v-html property which is possible when using router-link, but not with g-link.

### Steps to reproduce

Retrieve some data from an external API with special characters. Display them in a component with the v-html property set (this doesn't work and there's no error output on the console):

```
        <ul>
          <li v-for="{ node } in $static.posts.edges" :key="node.id">
            <g-link :to="node.path" v-html="node.title"></g-link>
          </li>
        </ul>
```

### Expected result
The output of the unescaped title.
### Actual result
The link text doesn't render.
### Environment

 System:
    OS: macOS 10.15.5
    CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
  Binaries:
    Node: 14.4.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 83.0.4103.116
    Firefox: 77.0.1
    Safari: 13.1.1

thetre97:
@mensch Don't think it is a bug exactly - both `g-link` and `router-link` have a slot where you should put content. So you can do something like this for example:

```vue
<g-link :to="node.path">
  <div v-html="node.title" />
</g-link>
```

mensch:
Right, I was overthinking it. So I added a span wrapped by g-link. Still, minor issue, but router-link does seem to support v-html directly without the need for wrapping the link text inside another element.

thetre97:
> but router-link does seem to support v-html directly

Yeh, okay - I guess as `g-link` wraps it, it doesn't pass that binding down.

CodingCarter:
> ## Description
[...]

PR #1373, opened 2020-10-24
fix(<g-link>): make sure that other domProps are properly passed to <router-link>

Erbenos:
Addresses #1269


hjvedvik:
Thanks again, @Erbenos :) Should I mark this one too?

Erbenos:
Would be awesome, thanks :)

PR #1405, opened 2020-12-24
fix(source-wordpress): non-paginated request on users endpoint causing truncated data

fmakdemir:
The users endpoint of wp-json api by default provides 10 users (or authors).
Since the previous version was using a request without pagination some posts
could get null author fields causing problems in template building.

This small change turns them into paged requests like some other types are already
doing.


hjvedvik:
Thanks, @fmakdemir! And sorry for this late reply. Your PR looks good to me :)

PR #1421, opened 2021-01-17
Fixes for tests under windows / jest update to 26.6.3

grbd:
Hi,
I've recently updated the version of jest and made some changes so that the tests will run under both windows and linux
before windows was throwing some errors due to unexpected results from the path delimiter and because of a slight difference in the file naming with webpack

I've tested under both windows 10 and Manjaro linux, also the yarn.lock file has been updated (I used the changes under the linux environment after a yarn install)


hjvedvik:
Thanks a lot for doing this, @grbd! Your changes look mostly good to me. I just have a few comments, so I'll do a proper review one of the next few days.

PR #1429, opened 2021-02-02
feat(pager): showLinks option show only prev/next links(#215)

### LINKED ISSUE #215: Feature request: Pager component: show only prev/next links
Looks like it's not possible to show only next/prev links with the Pager.

From my understanding, `showLinks` would be used to enable/disable the _numeric_ page links? Because if it's meant to completely remove the pagination, you just wouldn't use the component, right?

If I set `showLinks` to `false`, the entire navigation disappears.

### Steps to reproduce

Try using the Pager component like this:

```vue
<Pager :info="$page.myobject.pageInfo" :showLinks="false"/>
```

### Expected result

See only next/prev links in pagination (numeric page links removed).

### Actual result

Pagination is completely removed.

kyoruni:
Close #215
I wanted to use this option(#215), so I fixed it.

## Example

Pager component like this.

Option → `:showLinks="false"`

```vue
<Pager :info="$page.myobject.pageInfo" :showLinks="false"/>
```

## Result

See only next/prev links in pagination (numeric page links removed).

### DIFF
--- a/gridsome/app/components/Pager.js
+++ b/gridsome/app/components/Pager.js
@@ -78,7 +78,8 @@ export default {
       if (end < total) links.push(renderLink(total, lastLabel, ariaLastLabel, [lastClass, navClass]))
     }
 
-    if (links.length < 2) return null
+    if (showLinks && links.length < 2) return null
+    if (!showLinks && links.length === 0) return null
 
     return h('nav', {
       ...data,


hjvedvik:
Thanks, @kyoruni! Works perfectly :)

PR #1430, opened 2021-02-02
Add g-image fallback support

u12206050:
Keep displaying dataUri (placeholder) if loading the src results in error either due to file not found or being offline.


hjvedvik:
Thanks, @u12206050 :)

PR #1432, opened 2021-02-05
Add database formatted DateTime to GraphQL

kylewelsby:
Here I am adding support for Database formatted DateTime. 

Some of my imported data appear to be in this format and querying it via GraphQL yields an error, but `moment.js` can successfully parse the format. 

I hope this subtle change helps other projects facing the same Date type issue. 

example:

```
2011-02-25 23:59:59.999 +0530
```

### DIFF
--- a/gridsome/lib/graphql/types/date.js
+++ b/gridsome/lib/graphql/types/date.js
@@ -1,9 +1,9 @@
 const moment = require('moment')
 const { Kind } = require('graphql')
-const { ISO_8601_FORMAT } = require('../../utils/constants')
+const { SUPPORTED_DATE_FORMATS } = require('../../utils/constants')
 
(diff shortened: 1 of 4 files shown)


kylewelsby:
If you're falling into this issue and want a workaround to the `invalid date` issue that happens when your data is not exactly as expected. 

Try the following. 


```javascript
// gridsome.server.js
module.exports = function(api) {
  api.loadSource(({ getCollection }) => {
    const posts = getCollection("Post")
    posts.data().forEach((post) => {
      if (post.date) {
        post.date = moment.utc(post.date).toISOString()
      }
    })
  })
})
```

hjvedvik:
Thanks, @kylewelsby :) It seems like a good idea to support those formats too.

PR #1461, opened 2021-03-16
Server-Side Request Forgery

do-io:
https://www.npmjs.com/advisories/1594


hjvedvik:
Thanks, @do-io. Can you do these updates in the `master` branch instead? :)

PR #1465, opened 2021-03-24
Fix to Update CSS Loader to ^5.1.2

jvelez5239:
Updated CSS Loader to ^5.1.2 in order to suffice for other packages that require newer versions of this package


hjvedvik:
Thanks for the PR, @jvelez5239. Here is the error message from the failing pnp test:

```
ValidationError: Invalid options object. CSS Loader has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'localIdentName'. These properties are valid:
   object { url?, import?, modules?, sourceMap?, importLoaders?, esModule? }
```

PR #1470, opened 2021-03-31
Use webpack-dev-server for local dev

hjvedvik:
This PR uses `webpack-dev-server@4` as the local dev server.


RehanSaeed:
Cool!

AllanChain:
Seems `api.configureServer` hook is receiving `undefined`, should be an express `app`. That's because `server.app` is `undefined` until `initialize`

AllanChain:
A possible fix would be:
```patch
-  devServer.onBeforeSetupMiddleware = (server) => {
+  devServer.onBeforeSetupMiddleware = async (server) => {
     setupGraphQLMiddleware(app, server)
     setupAssetsMiddleware(app, server)
+    await app.plugins.configureServer(server.app)
     onBeforeSetupMiddleware && onBeforeSetupMiddleware(server)    
   }

   const server = new WebpackDevServer(compiler, devServer)        

-  await app.plugins.configureServer(server.app)
-
   return server
 ```

But it looks like that `webpack-dev-server` isn't `await`ing `onBeforeSetupMiddleware`, which may cause some problems.

PR #1477, opened 2021-04-09
feat(source-contentful): added entry metadata

adambateman:
This is a really simple one-line change to the source contentful plugin that will pull the in the `metadata` field which contains tags as part of Contentful's new [tagging feature](https://www.contentful.com/help/tags/). This allows you to tag assets to organise content, and we're using it to help determine what content should get built or ignored in staging/prod versus development.

Screenshot 2021-04-09 at 11 00 27

Here's a sample of what this looks like in GraphQL. 

Screenshot 2021-04-09 at 11 33 53

### DIFF
--- a/packages/source-contentful/index.js
+++ b/packages/source-contentful/index.js
@@ -105,7 +105,7 @@ class ContentfulSource {
       }
 
       node.id = entry.sys.id
-      node.metadata = entry.metadata;
+      node.metadata = entry.metadata
 
       collection.addNode(node)
     }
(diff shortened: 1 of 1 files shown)


noxify (approved):
LGTM

iampluque:
@adambateman @noxify will it be publish to npm !?

noxify:
@iampluque i don't think so - Maybe better to move to Nuxt or something else which is in active development 🙈

iampluque:
ok @noxify so gridsome is no longer in active development ?

noxify:
@iampluque  - to be honest, i don't know the current status.

my opinion: based on the latest changes/commits/releases/etc., yes i think it's currently not in active development.

@hjvedvik / @tomtev - Please correct me if I'm wrong, I don't want to say something which isn't true ;)

PR #1482, opened 2021-04-21
refactor: upgrade sharp for ~2x faster image processing

lovell:
Hello, I hope this small change can help make gridsome a bit more awesome.

The features of `imagemin-mozjpeg` and `imagemin-pngquant` are now directly available via `sharp` (the use of `imagemin-webp` is unnecessary as it produces the same output as the input).

- Removes imagemin decode/encode roundtrip per image, effectively doubling throughput and reducing peak memory usage
- Reduces `node_modules` filesystem size by ~10%, improving installation time by a similar amount
- Removes no longer required libxml 'warmup'
- Updates test snapshots that demonstrate smaller PNG output

Upgrading also provides support for M1 users.


hjvedvik:
Thank you so much, @lovell! This is a really great upgrade :)

PR #1486, opened 2021-04-28
chore(cli): return with an exit code of 1 if the target directory is non-empty

jamesgeorge007:
Currently, an attempt to create a project in a non-empty directory returns an exit code of `0`. Ideally, a failed event should return with a non-zero exit code. This PR aims at bringing about this behavior change.


3imed-jaberi (approved):
lgtm

Issue #869, opened 2019-12-09
Gridsome fails when using the source drupal plugin 


dasjo:
## Description

The gridsome develop step fails with the following error

Error: Type with name "DrupalTaxonomyVocabulary" does not exists

### Steps to reproduce

Install a drupal 8 site, enable json api
follow gridsome drupal steps from https://gridsome.org/plugins/@gridsome/source-drupal

dasjo@jos-macbook-pro my-gridsome-site % gridsome develop

dasjo@jos-macbook-pro my-gridsome-site % gridsome develop
Gridsome v0.7.11

Initializing plugins...
DrupalMenuLinkContent has no entities
DrupalImageStyle has no entities
DrupalCommentType has no entities
DrupalBlockContentBasic has no entities
DrupalComment has no entities
DrupalNodePage > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodePage > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodePage > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodePage > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodePage > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeExperience > Failed to add node: Duplicate key for property path: [object Object]
DrupalBlockContentType has no entities
DrupalTaxonomyVocabulary has no entities
Load sources - 0.16s
Error: Type with name "DrupalTaxonomyVocabulary" does not exists
    at SchemaComposer.get (/Users/dasjo/Drive/04_Projects/my-gridsome-site/node_modules/graphql-compose/lib/TypeStorage.js:44:13)
[...]

PR #1489, opened 2021-05-05
feat(drupal): handle path attribute in drupal's jsonapi response

superkumario64:
In Drupal's jsonapi response there is a `path` attribute, the value is an object with some attributes like `alias` see here: https://dev-campus-drupal-9.pantheonsite.io/jsonapi/node/article

When there are more then one article you see this when running `gridsome develop`:
```
DrupalNodePage > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodePage > Failed to add node: Duplicate key for property path: [object Object]
DrupalNodeArticle > Failed to add node: Duplicate key for property path: [object Object]
```

Only one node for that type will ever get loaded into GraphQL. In the output above I should have 3 DrupalNodePages and 2 DrupalNodeArticle. But when I go into the GraphQL explorer I only have 1 of each type.

Evidence of error is apparent in #869 

They had a permission's error that killed the build all together, however there is evidence of the duplicate key error.

===

Edit:

@matt-e-king was kind enough to hop on a Zoom call with me and we ran some tests on the development environment I had setup. We determined that due to Gridsome and Drupal having conflicting uses for the `path` attribute it would be best to rename Drupal's `path` to `drupal_path` when pulling it into the Gridsome data layer.

### DIFF
--- a/packages/source-drupal/README.md
+++ b/packages/source-drupal/README.md
@@ -93,6 +93,7 @@ user--user -> DrupalUser
 
 > The `keys` above are the defaults, but can be changed using [JSON:API
 > Extras](https://www.drupal.org/project/jsonapi_extras) (i.e. `node--article` can become just `article`
+> Note: JSON:API can return an attribute `path` that contains an object. This `path` attribute conflicts with Gridsome's GraphQL `path` field. To avoid conflict the Drupal data contained in `path` has been renamed to `drupal_path`.
 
 The url `values` of the `links` object get looped over and requested via `axios` and each response becomes processed, converted into `nodes` on each respective GraphQL Type. Each entity response comes with a `relationship` object which is merged into each `node` - the Gridsome Core then intelligently creates all the GraphQL Connects. [See example page queries below](#example-page-queries).
 
--- a/packages/source-drupal/lib/Entity.js
+++ b/packages/source-drupal/lib/Entity.js
@@ -109,7 +109,8 @@ class Entity {
 
   processFields (fields = {}) {
     const processedFields = reduce(fields, (newFields, field, key) => {
-      newFields[key] = (field !== 'null') ? field : ''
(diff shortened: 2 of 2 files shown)


matt-e-king:
@superkumario64 Hey there! You'll have to forgive me, I haven't used Gridsome or helped maintain this package in a LONG time. I believe I have a hunch on what is going on. Gridsome has the ability to generate a URL path based for each node processed to give you friendly URLs. This means each "node" already has a `path` attribute so if your response comes with a `path` attribute it will break.

Forgive me if I'm wrong, but your API response contains a `path` attribute because you are using the [Pathauto Drupal Module](https://www.drupal.org/project/pathauto), correct?

With Gridsome, you shouldn't need this Drupal module because Gridesome and the `source-drupal` plugin can generate path aliases for you. In your `gridsome.config.js` you can specify [templates](https://github.com/gridsome/gridsome/tree/master/packages/source-drupal#routing) to specify the URL schema for each Drupal Entity type.

I'm wondering (if you are using Pathauto) if you disabled that module in Drupal, and added a `template` entry for `DrupalNodePage` and `DrupalNodeArticle` if this would satisfy your needs? You can use any other attribute to create the URL allias.
```
templates: {
    DrupalNodePage: '/page/:title',
    DrupalNodeArticle: '/article/:name'
  }
```

I can see how your suggested fix in this PR would work, but not sure it's necessary. Let me know what you find out!

superkumario64:
@matt-e-king I wasn't using the https://www.drupal.org/project/pathauto plugin. It's a fresh install of Drupal 9 and by default it seems this core Drupal module is enabled: https://www.drupal.org/docs/8/core/modules/path

This was adding the path attribute, after I disabled it I was able to get all the content into graphql. The change in this PR may be nice since by default in a fresh Drupal install the JSONAPI has a `path` attribute that breaks the loading of Graphql.

matt-e-king (approved):

PR #1497, opened 2021-05-09
feat: afterSourcesLoaded hook

mklueh:
What I was trying to achieve was connecting collection entries of two sources with each other, based on certain fields.

However, using the `getCollection` action in the `loadSources` hook for this case won´t work, as the collections from the source plugins aren´t loaded before, but afterwards and getCollection always returns 'undefined'.

To solve this, I´ve created a new hook, called `afterSourcesLoaded` which is called after all source plugins have fetched their data and created their corresponding collections. You can do manipulation on all available data (that is loaded with the data store api).

Although it is working as expected, I´m not sure if it´s done correctly. For example I´ve used the BOOTSTRAP_SOURCES phase twice, not sure if it has side effects in some way.

Also, the createSchemaActions(api, this._app) is called from loadSources and afterSourcesLoaded. Consequences if any?

### DIFF
--- a/gridsome/lib/app/PluginAPI.js
+++ b/gridsome/lib/app/PluginAPI.js
@@ -49,6 +49,10 @@ class PluginAPI {
     this._on('loadSource', handler)
   }
 
+  afterSourcesLoaded(handler){
+    this._on('afterSourcesLoaded', handler)
+  }
+
   createSchema (handler) {
     this._on('createSchema', handler)
   }
--- a/gridsome/lib/app/PluginAPI.js
+++ b/gridsome/lib/app/PluginAPI.js
@@ -49,10 +49,6 @@ class PluginAPI {
     this._on('loadSource', handler)
   }
 
-  afterSourcesLoaded(handler){
-    this._on('afterSourcesLoaded', handler)
-  }
-
   createSchema (handler) {
     this._on('createSchema', handler)
   }
--- a/gridsome/lib/app/Plugins.js
+++ b/gridsome/lib/app/Plugins.js
@@ -26,8 +26,13 @@ class Plugins {
     )
 
     app.hooks.bootstrap.tapPromise(
-      { name: 'createSchema', label: 'Create GraphQL schema', phase: BOOTSTRAP_GRAPHQL },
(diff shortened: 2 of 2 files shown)


hjvedvik:
Thanks, @mklueh. Having a hook after the sources have loaded seems like a good idea :) Should we name it `afterLoadSource` instead, maybe? 

I'm not sure when we'll publish the next version yet, unfortunately. But in the meanwhile, you can try something like this in your `gridsome.server.js`:

```js
module.exports = (api) => {
  const { hooks, store } = api._app

  hooks.bootstrap.tapPromise({ name: 'YourHookName', before: 'createSchema' }, async () => {
    const collection = store.getCollection('YourCollection')
  })
}
```

mklueh:
> Thanks, @mklueh. Having a hook after the sources have loaded seems like a good idea :) Should we name it `afterLoadSource` instead, maybe?

Good point, but would you rather agree with `afterLoadSources` to keep naming consistency with `loadSources`?



> I'm not sure when we'll publish the next version yet, unfortunately. 

Is something blocking new releases currently? I feel like minor changes like this one should be able to be pushed to master more frequently and released more often.

> But in the meanwhile, you can try something like this in your `gridsome.server.js`:
> 
> ```js
> module.exports = (api) => {
>   const { hooks, store } = api._app
> 
>   hooks.bootstrap.tapPromise({ name: 'YourHookName', before: 'createSchema' }, async () => {
>     const collection = store.getCollection('YourCollection')
>   })
> }
> ```

That's good to know, I'll try that. Thank you :)

mklueh:
Hi @hjvedvik ,

This does unfortunately not work.
```

    hooks.bootstrap.tapPromise({name: 'afterLoadSources', before: 'createSchema'}, async () => {
        const collection = store.getCollection('Products')
        const post = store.getCollection('Post')

        console.log(collection)
        console.log(post)
    })

    api.afterLoadSources(({getCollection}) => {

    })

```

the hook is not available and I get `gridsome.server.js: api.afterLoadSources is not a function (25:9)`.

I´m also not sure how, if this would work the hook is called. Would that be done automatically because of the 'before' property?

PR #1504, opened 2021-05-15
feat: create json data hash on route basis

AllanChain:
*This is a draft PR and proof of concept. Tests not passed and code not cleaned. And I want to have some disscussion.*

Create json data hash based on route query source. e.g.

```
data
├── 404.json
├── a-post-with-a-cover-image
│   └── index.4b9dd3d1.json
├── index.e463b400.json
├── markdown-test-file
│   └── index.4b9dd3d1.json
├── say-hello-to-gridsome
│   └── index.4b9dd3d1.json
└── tag
    ├── Cover Image
    │   └── index.def7a3cd.json
    ├── Markdown
    │   └── index.def7a3cd.json
    ├── Releases
    │   └── index.def7a3cd.json
    └── Test files
        └── index.def7a3cd.json
```

Benifits:
- Keep data versioned by their structure
- Thus avoid most incapability issue when browser caches old data file
- In PWA, keeping old json data allows old version work properly before upgrade

This change tries to replace the hash generated by webpack, because currently
- This hash is somewhat inconsistent
- One single data change causing all html and json changing
- Thus makes incremental build impossible

Generating hashes per page, although it solves drawbacks of using webpack hash and has benifit of hashing route query, requires a large table to look up, which is impractical with large sites.

I think this feature should be controled by an option. It is suggested to work without hash-meta and hash in json data, but turning `cacheBusting` off has a side effect of js and css hashing. How to design the options?

### DIFF
--- a/gridsome/app/fetch.js
+++ b/gridsome/app/fetch.js
@@ -59,7 +59,7 @@ export default (route, options = {}) => {
 
   return new Promise((resolve, reject) => {
     const usePath = route.name === NOT_FOUND_NAME ? NOT_FOUND_PATH : route.path
-    const jsonPath = route.meta.dataPath || unslashEnd(usePath) + '/index.json'
+    const jsonPath = route.meta.dataPath || unslashEnd(usePath) + `/index.${route.meta.dataHash}.json`
     const absPath = unslashEnd(dataUrl) + jsonPath
 
     if (shouldPrefetch && !isLoaded[jsonPath]) {
--- a/gridsome/lib/app/build/createRenderQueue.js
+++ b/gridsome/lib/app/build/createRenderQueue.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const crypto = require('crypto')
 const { pathToFilePath } = require('../../pages/utils')
 const { trimEnd } = require('lodash')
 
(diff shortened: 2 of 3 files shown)


hjvedvik:
I think this could be a good idea, @AllanChain. One thing that can be a problem is that we are probably going to remove the static routes when we release a version for Vue 3. The routes will be added dynamically instead while browsing the site. The plan is to check for a JSON file at the same path as the page path, load it and create a route based on data in it (to load the correct component etc.).

But we should of course discuss if there are better alternatives to storing the webpack hash in each JSON file. One approach could be to store the webpack hash in a single file (for example `app.json`) which is fetched before each page JSON file. The `app.json` file would be cached by the browser, so fetching it multiple times shouldn't matter. The Etag header for `app.json` will tell if the webpack hash has changed and refresh the browser. I think this is roughly how it's done in Gatsby too.

AllanChain:
Yes, Gatsby's `app-data.json` is great enough for normal apps. And PWA will work too, if adding a hook when hashes don't match. (e.g. hook to send custom messages to service worker before reloading page, like https://github.com/gatsbyjs/gatsby/blob/8837033bf1353e9e8af9d9a7127e8534a9ddce4d/packages/gatsby/cache-dir/navigation.js#L93-L108)

However, there are still some cases uncovered with this approach in PWA, and I think keeping data versioned by query hash is a good way to avoid these problems

#### 1. app shell mode

In app shell mode, we first generate a "blank" (with js and css but without data or rendered HTML) page from a "blank" (`export default { render: () => null }`) vue file. Then we cache it and use it for all pages, fetching data according to the path. In this case, if cached js and new data conflicts, the page can't disply correctly.

#### 2. caching every page HTML

This way data are contained in page HTML, first page load succeeds. But when navigating and hash not match occurs, and if new service worker aren't ready, the next page will have problem.

---

The root of the problems is we have to use cached js and HTML to make page load fast in PWA, but cached js may conflict with new data. Gatsby has similar problems too (e.g. gatsbyjs/gatsby#13410). But I don't think their workaround is elegant. We should avoid page crash and preserve speedy nature of PWA, instead of reloading after crash. And if page doesn't crash and service worker takes some time to install, reloading page when user is browsing is an awful UX.

AllanChain:
As for the implementation, I think we can expose some API and let plugins change the name of `index.json`, and write the name as extra data to that "JSON file at the same path as the page path".

As the internal API may change after Vue 3 version, it's better to discuss after that release. So I am closing this now and waiting for Vue 3 version.

Issue #956, opened 2020-01-27
Markdown file title not support Chinese?


chenhongen:
## Description

(node:6268) UnhandledPromiseRejectionWarning: TypeError: Expected "title" to be a string

### Steps to reproduce

When set title with Chinese in markdown file,It show the error.

### Expected result

Support Chinese

### Actual result

![image](https://user-images.githubusercontent.com/16499303/73157312-905a9780-411b-11ea-935c-6c5f112447fc.png)

### Environment

gridsome: 0.7.12

tangyiming:
you can change the gridsome config. 
Add a custom path rather than use route.
```
// This is where project configuration and plugin options are located. 
// Learn more: https://gridsome.org/docs/config

// Changes here requires a server restart.
// To restart press CTRL + C in terminal and run `gridsome develop`

module.exports = {
  siteName: 'blog by Eamon',
  siteDescription: 'While the world sleep I dream',
  siteUrl: 'https://tangyiming.github.io',
  plugins: [
    {
      // Create posts from markdown files
      use: '@gridsome/source-filesystem',
      options: {
        typeName: 'Post',
        path: 'content/posts/*.md',
        // route: '/:slug',
        refs: {
          // Creates a GraphQL collection from 'tags' in front-matter and adds a reference.
          tags: {
            typeName: 'Tag',
            route: '/tag/:id',
            create: true
          }
        }
      }
    }
  ],

  transformers: {
    //Add markdown support to all file-system sources
    remark: {
      externalLinksTarget: '_blank',
      externalLinksRel: ['nofollow', 'noopener', 'noreferrer'],
      anchorClassName: 'icon icon-link',
      plugins: [
        '@gridsome/remark-prismjs'
      ]
    }
  },
}
```

zacnomore:
Closing as a duplicate of #918, the test cases in that one are very thorough!
I'm going to link to your workaround @tangyiming in there as well

chenhongen:
I tried your method, although it made it complicated, it didn't give me error again.By the way,Do you know how to make `gridsome-plugin-flexsearch` to surpport Chinese?

kongmoumou:
Currently my solution is to set `path` option as a function in the `templates` config.

This is my `gridsome.config.js`

```js
module.exports = {
  ...
  templates: {
    Post: [
      {
        path: (node) => {
          const postPathPrefix = /^\/content\/posts/;
          return node.path.replace(postPathPrefix, '');
        }
      },
    ],
[...]

Issue #918, opened 2020-01-05
Cannot Parse Korean in markdown file


perade:
## Description

I'm sorry, but I'm writing here because it seems to be a gridsome or @gridsome/source-filesystem problem.
I executed a [gridsome-starter-blog](https://github.com/gridsome/gridsome-starter-blog) in the local environment to see how gridsome works.
**No code or configure have been modified.**
I wanted to modify the post to Korean because I am Korean.
However, the error occurred when I typed it to Korean, and no post was displayed.
The same error occurs in other CJKs(Chinese, Japanese, Korean).

### Steps to reproduce

Modify the title of `gridsome-starter-blog\content\posts\*.md` file in path as shown below.
It only works if the title contains English or numbers.
_It doesn't work if there's only CJK or if there's CJK and a symbol together_.

```
// AS-IS (work)
---
title: Markdown test file
date: 2019-02-06
published: ...
---

// TO-BE 1 (not work)
---
title: type each word => 테스트 / 测试中 / テスト
date: 2019-02-06
published: ...
---

// TO-BE 2 (not work)
---
title: 테스트..
date: 2019-02-06
published: ...
---

// TO-BE 3 (work)
---
title: 테스트 123
date: 2019-02-06
published: ...
---

// TO-BE 4 (work)
---
title: 테스트 test
date: 2019-02-06
published: ...
---
```

### Expected result

Display post with Korean

### Actual result

Error occurred
```
(node:115948) UnhandledPromiseRejectionWarning: TypeError: Expected "title" to be a string
    at C:\Users\perade\GitHub\gridsome-starter-blog\node_modules\path-to-regexp\index.js:194:13
    at makePath (C:\Users\perade\GitHub\gridsome-starter-blog\node_modules\gridsome\lib\plugins\TemplatesPlugin.js:77:10)
    at C:\Users\perade\GitHub\gridsome-starter-blog\node_modules\gridsome\lib\plugins\TemplatesPlugin.js:205:24
    at SyncBailWaterfallHook.eval [as call] (eval at create (C:\Users\perade\GitHub\gridsome-starter-blog\node_modules\gridsome\node_modules\tapable\lib\HookCodeFactory.js:19:10), <anonymous>:19:16)
    at Collection.updateNode (C:\Users\perade\GitHub\gridsome-starter-blog\node_modules\gridsome\lib\store\Collection.js:127:50)
[...]

PR #1513, opened 2021-06-01
feat(makePath): use raw string as a fallback when the slugify result is empty

ceynri:
## Why

When we use CJK characters as the routing path, gridsome returns an error because slugify result is an empty string that cannot be used as the routing path.

Although the URL without slugify may not be beautiful, users generally prefer to build successfully first and then look for a solution likes to add `slug` in the markdown file or use `:fileInfo__name` as route path if they are not satisfied with the encoded URL. (Relate issues #918, #956)

## My test

```
// case 1 (raw title fallback)
title: '테스트'
result: http://example.com/%ED%85%8C%EC%8A%A4%ED%8A%B8/

// case 2 (raw title fallback)
title: '中文测试'
result: http://example.com/%E4%B8%AD%E6%96%87%E6%B5%8B%E8%AF%95/

// case 3 (affected by slugify)
title: 'test 中文测试'
result: http://example.com/test/

// case 4
title: 'slugify() test'
result: http://example.com/slugify-test/
```

### DIFF
--- a/gridsome/lib/plugins/TemplatesPlugin.js
+++ b/gridsome/lib/plugins/TemplatesPlugin.js
@@ -65,9 +65,12 @@ const makePath = (object, { path, routeKeys, createPath }, dateField = 'date', s
         ) {
           return String(value.id)
         } else if (!isPlainObject(value)) {
-          return suffix === 'raw'
-            ? String(value)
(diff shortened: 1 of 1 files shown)


noxify (approved):
LGTM

PR #1523, opened 2021-06-28
Upgrade to webpack 5

hjvedvik:
This PR upgrades webpack to v5.


tyankatsu0105:
Wow!! Awesome works!!

OzzyTheGiant:
Ideally I would like this to come soon as I want to upgrade PostCSS to 8.0 so that I can use TailwindCSS 3.0, but I can't do that because PostCSS 8.0 needs webpack 5

PR #1550, opened 2021-08-19
Drop support of node 13 and update related docs

kecrily:
- Format markdown files
- Add way to install with pnpm
- Modify the development requirements mentioned in README.md that forgot to change in https://github.com/gridsome/gridsome/commit/10c5ac1eef96e6c7f62f5be86dbf468a21d7d043
- Drop support of node 13 (end in 2020-06-01)

---

node15 also ended its life on June 1st, do we need to drop it too?


hjvedvik:
Thanks for looking into this, @kecrily :) The changes look very good to me! And I think it should be okay to drop support for node 15 too. So feel free to update the PR or else I'll merge it as it is now.

kecrily:
Now node 15 was also dropped

PR #1586, opened 2021-11-17
Fix transformer-remark visit

jeffpyke:
The return value matters for the visitor function. This bug was causing some images and links to be skipped when parsing the remark AST. The result was that images would show up as `img` with a invalid src value instead of `g-image`.

https://github.com/syntax-tree/unist-util-visit-parents#returns


hjvedvik:
Thanks, @jeffpyke :)

Issue #1590, opened 2021-12-03
Use @gridsome\source-strapi: TypeError: docs.forEach is not a function


Suwings:
## Description

I want to build a CMS using Gridsome and Strapi. When I use the @gridsome/source-strapi plug-in, the data cannot be imported.

The error is as follows:

```
PS C:\Workspace\my-gridsome-site> gridsome develop
Gridsome v0.7.23

Initializing plugins...
Fetching data from Strapi (http://localhost:1337/api)
TypeError: docs.forEach is not a function
    at C:\Workspace\my-gridsome-site\node_modules\_@gridsome_source-strapi@0.2.0@@gridsome\source-strapi\index.js:40:30
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Promise.all (index 0)
    at async Promise.all (index 0)
    at async Plugins.run (C:\Workspace\my-gridsome-site\node_modules\_gridsome@0.7.23@gridsome\lib\app\Plugins.js:141:11)
```

Strapi has been launched.
visit http://127.0.0.1:1337/api/blog-posts
```json
{"data":[{"id":1,"attributes":{"title":"TestArticle2","createdAt":"2021-12-03T02:11:03.034Z","updatedAt":"2021-12-03T03:10:21.841Z","publishedAt":"2021-12-03T02:11:36.622Z","date":"2021/1/1","text":"sssssssssssssssssss"}},{"id":2,"attributes":{"title":"TestArticle1","createdAt":"2021-12-03T02:11:15.365Z","updatedAt":"2021-12-03T03:10:14.606Z","publishedAt":"2021-12-03T02:11:33.755Z","date":"2021/1/1","text":"sfooooooooo"}},{"id":3,"attributes":{"title":"TestArticle3","createdAt":"2021-12-03T02:11:26.609Z","updatedAt":"2021-12-03T03:10:29.575Z","publishedAt":"2021-12-03T02:11:39.189Z","date":"2021/1/3","text":"xxxxxxxxxxxxxxxxxxxxxxxxx"}}],"meta":{"pagination":{"page":1,"pageSize":25,"pageCount":1,"total":3}}}
```

And when I use the following configuration (from the documentation):
```
module.exports = {
  siteName: 'Gridsome',
  plugins: [
    {
      use: '@gridsome/source-strapi',
      options: {
        apiURL: `http://localhost:1337/`,
        queryLimit: 1000, // Defaults to 100
        contentTypes: [`blog-posts`],
      },
[...]

PR #1599, opened 2022-01-09
fix error on build with source-strapi plugin

aldvine:
Hello. 
I have the same error of [the issue #1590](https://github.com/gridsome/gridsome/issues/1590)

The error: 
```
Initializing plugins...
Fetching data from Strapi (http://localhost:1337/api)
TypeError: docs.forEach is not a function
```

The plugin iterates with forEach on docs but it is an object type.
It is resolved for me with an "for in" iteration.


Okharev:
I have the same issue, this change seems to fix it

Issue #1545, opened 2021-08-15
Change package manager from yarn to pnpm


kecrily:
## Summary

Change package manager from yarn to pnpm.

### Motivation

From [the benchmark provided by pnpm](https://pnpm.io/benchmarks), it is easy to see that pnpm has better performance than yarn.

Speeding up the installation of dependencies can go a long way in improving the development experience.

So it should not be hard to make the decision to switch.

PR #1604, opened 2022-01-20
chore(pkg): replace yarn with pnpm

kecrily:
close https://github.com/gridsome/gridsome/issues/1545


kecrily:
I looked at the relevant code and I noticed that they hadn't been changed in a long time.

https://github.com/gridsome/gridsome/blob/149dbff92d8ea612553e7479991f2b43e4a903dd/packages/cli/__tests__/gridsome.spec.js#L8

https://github.com/gridsome/gridsome/blob/149dbff92d8ea612553e7479991f2b43e4a903dd/packages/cli/lib/utils/version.js#L6

According to the code in these two places, it is normal that this CI did failed. So how did the previous CI pass? I can't understand

```
    Expected pattern: /gridsome v(\d+\.?){3}/
    Received string:  "@gridsome/cli v0.3.4"

      13 |   const { stdout } = await runCLI(['-v'], { cwd: testPath })
      14 |
    > 15 |   expect(stdout).toMatch(/gridsome v(\d+\.?){3}/)
         |                  ^
      16 | })
      17 |
      18 | test('warn about unknown command', async () => {

      at Object.<anonymous> (packages/cli/__tests__/gridsome.spec.js:15:18)
```

According to the output Expected pattern and Received string, they can never match

hjvedvik:
Thanks for doing this @kecrily :)

The `gridsome -v` command currently outputs versions for `@gridsome/cli` and `gridsome` if you run it in a project that has `gridsome` as a dependency. So you could look into if it somehow isn't able to locate `gridsome` [here](https://github.com/kecrily/gridsome/blob/feature/pnpm/packages/cli/lib/utils/version.js#L46-L51).

PR #1648, opened 2022-08-19
automatically cache busting the json files based on the hash in meta tag

atflick:
Problem: json files get cached and when they are fetched their hash doesn't match the hash that gridsome generated in the HTML that's fetching the json files.  This results in the `Hash did not match` error which then breaks all pages until the user clears their cache.

This change simply adds the matching hash as a query param when the json files are fetched, meaning they can still be cached until that hash changes in which they should be freshly fetched.

This is the hash being added in the network tab:
image


atflick:
Solves this issue: https://github.com/gridsome/gridsome/issues/1032

ai-slave:
@hjvedvik can you please merge this one?

atflick:
bumping this again @hjvedvik 

This would greatly increase the confidence I have when deploying new changes to our public site.  We recently had a situation where there were a bunch of updates made to our site that got deployed earlier than we wanted and we rolled them back quickly but the json files were getting cached on the client side which was completely breaking the site for any users who had visited the newly updated version, then the rolled back version.  Similar situation happens on consecutive quick changes/updates

atflick:
@hjvedvik bumping this again, this simple change would help anyone still using gridsome out A LOT.

sergeyfilimonov:
@atflick how is this code working? there is no `hash` variable defined in the file so the build goes `hash is undefined`. So for now I've just changed `/index.json` in the file to `/index.json?h=Math.floor(Math.random() * 1000000000000000)`

atflick:
@sergeyfilimonov That works too, but the `hash` variable should be getting defined just above on line 57. Maybe it doesn't get defined in certain situations.  It's been a while since I looked at this.

image

PR #1701, opened 2026-05-13
feat: add markdown TOC and custom containers

### LINKED ISSUE #61: More Markdown extensions
Could we add more markdown extensions to the remark transformer plugin?

VuePress has some nice ones:
https://vuepress.vuejs.org/guide/markdown.html

Thinking about these two:
- Table of contents `[[toc]]`
- Custom containers `::: tip :::` etc.

landeqiming666:
## Summary

Fixes #61.

This adds built-in support to `@gridsome/transformer-remark` for the VuePress-style Markdown extensions requested in the issue:

- `[[toc]]` expands to a heading link list using the existing generated heading slugs
- `::: tip`, `::: warning`, `::: danger`, and `::: details` render as custom container blocks
- both features can be disabled with `tableOfContents: false` or `customBlocks: false`

The implementation keeps this inside the existing remark transformer pipeline and does not introduce a new Markdown parser.

## Notes

`remark-html` sanitization remains enabled. This PR only extends the default sanitize schema to preserve `className` values emitted by remark plugins, which is needed for the custom container classes.

## Verification

- `npx jest packages/transformer-remark/__tests__/index.spec.js --runInBand`
- `npm run lint -- packages/transformer-remark/**/*.js`
- `git diff --check`


landeqiming666:
Quick status update: this PR is still mergeable and keeps the Markdown extension support scoped to the existing remark transformer pipeline. The targeted transformer tests, lint for the changed package, and git diff check are listed in the PR body. Happy to adjust the scope or defaults if you prefer a smaller first pass.

